PR #11, opened 2020-07-23
Create presubmit.yml

freliciatucker:
- runs java , javascript and other format checkers as well as testing
- copied for resource jie suggested: https://github.com/googleinterns/step7-2020/blob/master/.github/workflows/presubmit.yml


.github/workflows/presubmit.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: Install ESLint with Google Config
+      run: npm install eslint@^7.3.1 eslint-config-google@^0.14.0
+    - name: Check
+      run: npx eslint src
```
zjwoody, line 16: Seems we don't have "src" directory, so the linter can't find the files, consider adjust the directory structure.

.github/workflows/presubmit.yml
```diff
@@ -0,0 +1,54 @@
...
+  lint-java:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Check
+      run: mvn spotless:check
```
zjwoody, line 27: You need add spotless plugin:
 
            <plugin>
                <groupId>com.diffplug.spotless</groupId>
                <artifactId>spotless-maven-plugin</artifactId>
                <version>1.31.3</version>
                <configuration>
                    <java>
                        <googleJavaFormat>
                            <version>${google-java-format.version}</version>
                        </googleJavaFormat>
                    </java>
                </configuration>
            </plugin>

zjwoody:
Hi Frelicia,
Could you try to fix the lintjs error? Seems it related to the directory structure.

.github/workflows/presubmit.yml
freliciatucker: sorry for the late response! . on the repo main page I see the "src" , so i confused on how to solve this problem

D-moe: This isn't an issue with the directory structure as the error would seem to indicate, it's returning an error as there are no javascript files in this directory. Adding an empty javascript file will resolve this issue. It might be better to change the path to src/main/webapp as this will reduce the search space for the command.

D-moe: For reference one can use this tool https://github.com/nektos/act along with installing docker to test these github actions locally. That is how I was able to discover this was the issue/

D-moe:
For configuring eslint I ran the tool locally and set up the config that way. The other project we were consulting has a slightly different config file so if we are not happy with this configuration we can change to that one, I just tried to pick the options that seemed most relevant to our specific application.

PR #13, opened 2020-07-28
Add protobuf encoding/decoding

D-moe:
Added the protobuf libraries to use for encoding .txt to .sstable format. Simple tests to check functionality. Update pom.xml file.


D-moe:
I'm not a huge fan of the look of code after running this java autoformatter, any thoughts on this, I mean I guess it's fine, but some of the stuff mvn:spotless apply looks a little unnatural to me?

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,24 @@
+package decoder;
+
+import com.google.protobuf.util.JsonFormat;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import proto.DataFormats.SerializedData;
+
+public class Decoder {
```
freliciatucker, line 9: can you add class comments?

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    Decoder.decode(input, output);
+    try {
+      input.close();
+      output.close();
+    } catch (IOException e) {
+      System.out.println("Unable to close the stream");
+    }
+
+    // Check to make sure that the output file exists.
+    File verifyFile = new File(outputFile);
```
freliciatucker, line 64: would there be any output file if user uploads no file or an empty file?

src/test/java/EncodeTest.java
```diff
@@ -0,0 +1,70 @@
...
+  private static OutputStream output;
+
+  @Before
+  public void setUp() {
+    inputFile = "testInput.txt";
+    outputFile = "testOutput.sstable";
+    try {
+      input = new FileInputStream(inputFile);
+    } catch (FileNotFoundException e) {
+      System.out.format("Unable to open the input file with name %s", inputFile);
```
freliciatucker, line 30: can the error message be more descriptive? like is it unable to print bc the given files the wrong format or bc the encoder is broken?

src/test/java/EncodeTest.java
```diff
@@ -0,0 +1,70 @@
...
+      System.out.println("Unable to close the stream");
+    }
+
+    // Validate the outputFile was generated
+    File verifyFile = new File(outputFile);
+    Assert.assertEquals(verifyFile.exists(), true);
+
+    // Validate that the size of the output file is smaller than the size of the
+    // input one (this should be true as protobuf compresses data at a smaller
+    // granularity than the txt encoding).
```
freliciatucker, line 65: just a question not necessarily a corrrection: if pictures are already compressed and we try to compress them more, the file sie may get a lil bigger bc it stores some extra overhead. is there any chance this could happen in our case? i remember Jie mentioning pose data could be uploaded as an image

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    } catch (SecurityException e) {
+      System.out.format("Permission error when opening the output file %s", outputFile);
+      e.printStackTrace();
+    }
+  }
+
+  @Test
+  /**
+   * Convert the encoded .sstable file (proto format) into json. Basic sanity checks, up to the user
+   * to make sure the output is indeed as expected.
```
freliciatucker, line 52: hi lol, this might be a dumb question, but what do you mean by sanity checks?

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+@RunWith(JUnit4.class)
+public final class DecodeTest {
+  private static String inputFile;
+  private static String outputFile;
+  private static InputStream input;
+  private static OutputStream output;
+
+  @Before
+  public void setUp() {
+    inputFile = "sampleProto.sstable";
```
freliciatucker, line 25: just to clarify, is all protobuf files in .sstable format?

freliciatucker:
Hey David, great work! many of my comments are just questions asking about clarification about how different parts work together, and some suggestions might can be implemented in later iterations of the product, not necessarily a priority. for readability purposes, could you add descriptive comments to your classes, functions, global variables, etc? i think also adding an empty file for unit testing could be a good test. and to respond to your comment on format, nothing looked too off for me!

target/maven-archiver/pom.properties
```diff
@@ -1,5 +0,0 @@
-#Generated by Maven
-#Tue Jul 21 14:32:08 UTC 2020
-version=1
```
freliciatucker, line 3: should this file be deleted?

src/test/java/EncodeTest.java
D-moe: I'm not totally sure what would be more descriptive, a filenotfoundexception occurs when the file cannot be found usually because it doesn't exist in the current diretory?

src/test/java/EncodeTest.java
D-moe: This implementation of the encoder currently only supports tesxt pose data upload, perhaps in the future the test can be modified slightly.

src/test/java/DecodeTest.java
D-moe: Just basic checks to make sure something reasonable is happening. Like we don't know if the json file is correct, but we at least know it is smaller than the original file and exists.

src/test/java/DecodeTest.java
D-moe: I believe that's the extension, it doesn't really matter as its just raw binary data

src/test/java/DecodeTest.java
D-moe: It's not the best check as this will always create a file as long as the file could be created I'll add another check to make sure the file is nonempty, as we know our test file is nonempty

src/test/java/EncodeTest.java
freliciatucker: oh ok then. i thought bout FileNotFoundException error messages were the same, but i notice that ones for input and the others for output

target/maven-archiver/pom.properties
freliciatucker: yea this happens to me too, do you know why target keeps making new files?

D-moe: when mvn builds it creates a .war file and other files to help it not relocate dependencies etc each time

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,24 @@
...
+  public static void decode(InputStream dataStream, OutputStream outputStream) {
+
+    // At some point it would make sense to have this data read in chunks of the
```
zjwoody, line 13: You can add a TODO here.

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.google.protobuf.util.JsonFormat;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import proto.DataFormats.SerializedData;
+
+public class Decoder {
+  public static void decode(InputStream dataStream, OutputStream outputStream) {
+
```
zjwoody, line 12: Unnecessary blank line.

src/main/java/proto/objects.proto
```diff
@@ -0,0 +1,30 @@
+/*
```
zjwoody, line 1: You can try to make the proto definition in a separate pr.

src/main/java/proto/objects.proto
```diff
@@ -0,0 +1,30 @@
...
+Instead of encoding all data in one proto, we will instead preface the proto
```
zjwoody, line 2: For the multi-line comments in header, maybe all use //, or /* and every line start with a *

I'd suggest  to use unified comment style.

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Validate that decoding the protobuf file to json works properly. */
```
zjwoody, line 15: Use //.
Or start a new line.

src/main/java/proto/objects.proto
```diff
@@ -0,0 +1,30 @@
...
+with a header field containing the number of total bytes in the proto. This will
+be a configurable value that will be computed where the biggest possible Data
+division below the maximum threshold is used. This will allow for easy chunking
+when reading in messages etc.
+*/
+syntax = "proto3";
+
+package formats;
+option java_package = "proto";
+option java_outer_classname = "DataFormats";
```
zjwoody, line 12: How about name it SensorData. since DataFormats doens't give an intuitive describe of what the proto is used to.
The same as the other naming, they are too general, try to give descriptive name.

src/main/java/proto/objects.proto
```diff
@@ -0,0 +1,30 @@
...
+
+message SerializedData { repeated Data data = 1; }
+message Data {
+  Position pose = 1;
```
zjwoody, line 16: The pose is 6D, it containts position and orientation.
message Pose {
double lat = 1;
double lng = 2;
double alt = 3;
float roll_deg = 4;
float pitch_deg = 5;
float yaw_deg = 6;
}
And describe each field, what your xyz axis is.

src/main/java/proto/objects.proto
```diff
@@ -0,0 +1,30 @@
...
+division below the maximum threshold is used. This will allow for easy chunking
+when reading in messages etc.
+*/
+syntax = "proto3";
+
+package formats;
+option java_package = "proto";
+option java_outer_classname = "DataFormats";
+
+message SerializedData { repeated Data data = 1; }
```
zjwoody, line 14: Do you need this proto only containts repeated Pose data?
For sstable, each record is a <timestamp, pose>, you don't need this proto of the repeated pose.

src/main/java/proto/objects.proto
zjwoody, line 1: The proto file name(objects.proto) is not meaningful.  you can call it Pose.proto since it defines the pose proto.
Try to give descriptive file name and class name instead of generic name(usually the file name is the same as class name so it is obviously for users to know what the file is).

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,25 @@
+package decoder;
+
+import com.google.protobuf.util.JsonFormat;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import proto.DataFormats.SerializedData;
+
+/** Decode data from the proto format to JSON. */
```
freliciatucker, line 9: is it mad bc of this extra star? I know im reaching, but cant think of why its upset

src/test/java/DecodeTest.java
D-moe: The autoformatter makes this the style. I agree that it doesn't look nice. I originally had it the way you suggested here.

zjwoody:
What's the testInput.txt is? It is too large so not be able to show. If it is only used in your unit test, you don't need such a big file. Only few lines of fake data is enough.
BTW, avoid checking in real data file into repo. it should only contains code.

D-moe:
testInput.txt is one of the sample inputs you gave us.

D-moe:
Shouldn't we check in real data for a test, I thought this made sense as Andre needs to use the generated json file, so it made sense to create it in the test?

zjwoody:
Try to split proto definition in a separate pr, and the encoder and decoder in another pr depend on proto pr.
Make each pr only constrains one feature.

D-moe:
You can't write the encoder/decoder without knowing the proto though right; wouldn't that make them the same feature?

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,25 @@
...
+public class Decoder {
+  public static void decode(InputStream dataStream, OutputStream outputStream) {
+
+    // At some point it would make sense to have this data read in chunks of the
+    // data file and include headers in the proto definitions so we could
+    // address issues with large files, stuff not fitting into RAM, etc.
+    try {
+      SerializedData protoData = SerializedData.parseFrom(dataStream);
```
zjwoody, line 17: I think here your goal is decode one record each time instead of the whole file at one time(some data will be very big that you are not able to store them all in memory), 
then you may need Frelica provide you an stream writer like DatabaseWriter->WriteProto(timestamp, PoseProto);  so you can write each piece of data into db.

freliciatucker: when you say one record, are you talking about decoding 1 point's location and orientation?

src/main/java/proto/objects.proto
D-moe: The original example I was going off uses a format similar to this, I think you are correct that this is unnecessary though I'm curious what the generated json will look like

src/main/java/proto/objects.proto
D-moe: Could you elaborate what you mean describe each field, I think I'm not fluent enough in the terminology here to know what you have in mind?

src/main/java/decoder/Decoder.java
zjwoody: Yes, each pose also has a timestamp, so one record is a pair of <timestamp, pose>.

zjwoody:
> You can't write the encoder/decoder without knowing the proto though right; wouldn't that make them the same feature?

So the pr 1 is the proto definition, 
Create pr 2 based on pr 1, it will contains all the code in pr1, and add your encoder/decoder in it.
Get review and check in the pr 1, then merge with pr 2, then check in pr 2.
This is a common way in development, split big pr into smaller one, make it can be reviewed and checked in asap.
Otherwise it will block the others who only need to know the proto definition.

zjwoody:
> Shouldn't we check in real data for a test, I thought this made sense as Andre needs to use the generated json file, so it made sense to create it in the test?

Never check in real data into code repo.
The goal of test is to make sure your implementation works as expectation. You should create fake data to test instead of using the real data. (e.g. you can create a simple pose trajectory by giving a short sequence of lat,lng,alt,roll,pitch,yaw, these data doesn't need to match with read world, you only need to test that your encoder/decoder can read txt and convert between different formats, and the actual value isn't change and missing during conversion, right?)

src/main/java/proto/objects.proto
zjwoody: For json, it is another story, not related to your proto definition. 
I think the json will be like this:
pose_data = [
{"timestamp": 123456, "pose":[1, 2, 3, 10, 0, 0]},
{"timestamp": 123457, "pose":[1, 2, 4, 10, 0, 0]},
{"timestamp": 123458, "pose":[1, 2, 5, 10, 0, 0]},
];

src/main/java/proto/objects.proto
zjwoody: Add comment for the field meaning in your proto definition, e.g.:
message Pose {
// Position is in WGS84 latitude, longitude, altitude.
double lat = 1;  // Degrees in [-90, 90].
double lng = 2; // Degrees in [-180, 180).
double alt = 3;  // Meters over WGS84 datum.
// The orientation, what's the local pose frame? what's the x,y,z direction? what's the angular unit?
float roll_deg = 4;
float pitch_deg = 5;
float yaw_deg = 6;
}

zjwoody: I think you need "optional", and you can also give the field a default value.

src/main/java/proto/objects.proto
D-moe: The build in json converter is based on the proto definition, so assuming we don't write the converter from scratch it would depend on the proto definition.

zjwoody: You can use json converter convert one pose data, and added it in your json array.
You'd better not add all data in one proto message, it will be very big, proto message is designed for small data, not you whole run's data.

D-moe:
No need to rereview yet, just pushed so I could show Frelicia some changes, still need to fix the decoder file.

src/main/java/decoder/Decoder.java
```diff
@@ -14,16 +14,19 @@ public static void decode(InputStream dataStream, OutputStream outputStream) {
     // data file and include headers in the proto definitions so we could
     // address issues with large files, stuff not fitting into RAM, etc.
     try {
```
freliciatucker, line 16: so once we update the reader and writer API's, is anything in the try necessary? is the goal to condense this function to only calling a read method and outputing it as json?

src/main/java/decoder/Decoder.java
```diff
@@ -14,16 +14,19 @@ public static void decode(InputStream dataStream, OutputStream outputStream) {
     // data file and include headers in the proto definitions so we could
```
freliciatucker, line 14: would the parameters change from taking a input and output stream to taking a reference for the file wanted from storage?

src/main/java/proto/Pose.proto
```diff
@@ -0,0 +1,22 @@
+/*
+Instead of encoding all data in one proto, we will instead preface the proto
+with a header field containing the number of total bytes in the proto. This will
+be a configurable value that will be computed where the biggest possible Data
+division below the maximum threshold is used. This will allow for easy chunking
+when reading in messages etc.
+*/
```
zjwoody, line 7: Block comment style:
subsequent lines must start with * aligned with the * on the previous line.
https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

src/main/java/proto/Pose.proto
```diff
@@ -0,0 +1,22 @@
...
+
+message Pose{
+  double gps_timestamp = 1;
+  double lat = 2;
+  double lng = 3;
+  double alt = 4;
+  float roll_deg = 5;
+  float pitch_deg = 6;
+  float yaw_deg = 7;
+}
```
zjwoody, line 22: Please add comment for each field.

src/main/java/proto/Pose.proto
```diff
@@ -0,0 +1,22 @@
+/*
+Instead of encoding all data in one proto, we will instead preface the proto
+with a header field containing the number of total bytes in the proto. This will
+be a configurable value that will be computed where the biggest possible Data
+division below the maximum threshold is used. This will allow for easy chunking
+when reading in messages etc.
```
zjwoody, line 6: The file comment should describe the proto, what it is used for.
I'm confused why you need a header field containing the number of total bytes in the proto?

src/test/java/EncodeTest.java
```diff
@@ -0,0 +1,70 @@
...
+/** Validate that a .txt file can be properly encoded into a .proto file. */
+public final class EncodeTest {
+  private static String inputFile;
+  private static String outputFile;
+  private static InputStream input;
+  private static OutputStream output;
+
+  @Before
+  public void setUp() {
+    inputFile = "testInput2.txt";
```
zjwoody, line 25: Rename the test file to give it a meaningful name, e.g. pose.txt

testInput2.txt
```diff
@@ -0,0 +1,4 @@
+gps_timestamp lat lng alt roll_deg pitch_deg yaw_deg
```
zjwoody, line 1: You may need move the test file into a separate directory, e.g. test/data/

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    } catch (SecurityException e) {
+      System.out.format("Permission error when opening the output file %s", outputFile);
+      e.printStackTrace();
+    }
+  }
+
+  @Test
+  /**
+   * Convert the encoded .sstable file (proto format) into json. Basic sanity checks, up to the user
```
zjwoody, line 51: What is "Basic sanity checks, up to user" means?
The unit test should have clear test goal.

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    } catch (IOException e) {
+      System.out.println("Unable to close the stream");
+    }
+
+    // Check to make sure that the output file exists.
+    File verifyFile = new File(outputFile);
+    Assert.assertEquals(verifyFile.exists(), true);
+
+    // Check to make sure that the output JSON file is large than the compressed
+    // protobuf one.
```
zjwoody, line 68: I think instead of check file size, you should check total record size and each filed value is as expectation.

src/test/java/EncodeTest.java
```diff
@@ -0,0 +1,70 @@
...
+      System.out.println("Unable to close the stream");
+    }
+
+    // Validate the outputFile was generated
+    File verifyFile = new File(outputFile);
+    Assert.assertEquals(verifyFile.exists(), true);
+
+    // Validate that the size of the output file is smaller than the size of the
+    // input one (this should be true as protobuf compresses data at a smaller
+    // granularity than the txt encoding).
```
zjwoody, line 65: ditto, check the record size and field value.

testInput2.txt
```diff
@@ -0,0 +1,4 @@
...
+1140885738.407125 48.129872 11.582905 10.015034 -177.971817 -3.228588 10.700191
```
zjwoody, line 2: Actually, you don't need this file, you can generate a temp file used to test. That is convenient for you to test the field value and record size.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private static String[] parseFields(String currentLine, String regex, int maxFields) {
+    // Split all the files based on a RegEx.
+    // TODO(morleyd): Add some file sanity checking, throw an error on
+    // unexpected format (might not do though as slows down operation).
+    String[] fields = currentLine.split(regex, maxFields);
+    return fields;
+  }
+
+  private static Pose parsePose(String[] input) {
```
zjwoody, line 73: Maybe you can make it as a lambda function and add it as a parameter of the encode funcion.

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,34 @@
+package decoder;
+
+import com.google.protobuf.util.JsonFormat;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import proto.SensorData.Pose;
+
+/** Decode data from the proto format to JSON. */
+public class Decoder {
```
zjwoody, line 10: You can make it a generic class.

D-moe: I think this is a good point, but would likely sense to implement in a later pull request. For MVP I think the current implementation should be sufficient.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,105 @@
...
+    Pose.Builder pose = Pose.newBuilder();
+    List<FieldDescriptor> currentFields = pose.getDescriptor().getFields();
+    ArrayList<Lambda> headingTypes =
+        new ArrayList<>(
+            Arrays.asList(toDouble, toDouble, toDouble, toDouble, toFloat, toFloat, toFloat));
+    setFields(pose, currentFields, input, headingTypes, 0, input.length);
```
jmchristen, line 79: This may look more elegant than setting the fields directly, but I don't think it's actually an appropriate place to reach for reflection. Reflection will almost always be slower and harder to understand, so it's better to avoid it unless you really don't know anything about the proto in question.

I'd recommend being explicit:
```
Pose pose = Pose.newBuilder()
  .setLat(currentFields[0])
  .setLng(currentFields[1])
  ...
  .build();
```

It means you won't have a reusable `setFields` method. You'll have to write similar methods for each proto that you eventually support. But that is well worth the increase in clarity and performance.

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    try {
+      output = new FileOutputStream(outputFile);
+    } catch (FileNotFoundException e) {
+      System.out.format("Unable to open the output file with name %s", outputFile);
+      e.printStackTrace();
+
+    } catch (SecurityException e) {
+      System.out.format("Permission error when opening the output file %s", outputFile);
+      e.printStackTrace();
+    }
```
jmchristen, line 46: I doubt all this exception handling makes much difference in the debugging experience when something fails. I'd recommend just letting the exceptions pass uncaught in a `setUp` method.

zjwoody:
Hi David, could you try to fix the comments and checkin the pr ?

D-moe:
I'll add these changes by end of the day, working on deconstructing that other branch into pull requests to merge in as well.

D-moe:
Any ideas why the autochecks didn't run here?
I believe this addresses all comments

D-moe:
I ended up changing the github action to run on push instead of pull_request which seemed to resolve the issue. I'm not sure why this issue just occurred now though?

src/test/java/DecodeTest.java
```diff
@@ -15,59 +21,92 @@
...
+  private static OutputStream decoderOutput;
+  private static String[] testInputArray;
+  private static String testInput;
+  private static String decoderInputFile;
 
   @Before
-  public void setUp() {
-    inputFile = "sampleProto.sstable";
-    outputFile = "testOutput.json";
+  public void setUp() throws SecurityException, FileNotFoundException {
```
jmchristen, line 34: For tests, you can just write `throws Exception`. No one will write code to call `setUp`, so there's no need to identify the specific exceptions. Same for the test cases below.

src/main/java/decoder/Decoder.java
```diff
@@ -0,0 +1,34 @@
...
+      while (currentPose != null) {
+        output.write(JsonFormat.printer().print(currentPose));
+        currentPose = Pose.parseDelimitedFrom(dataStream);
+        // Write ',' on all but last entry.
+        if (currentPose != null) output.write(",");
+      }
+      output.write("]");
+      // Close the PrintWriter stream.
+      output.close();
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
jmchristen: This comment still applies.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static void closeStream(OutputStream outStream) {
+    try {
+      outStream.close();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  private static String[] parseFields(String currentLine, int maxFields, String regex) {
```
jmchristen, line 56: This overload is never used except by the other overload. You can delete it and just do the `split` in the other overload.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,80 @@
...
+      // The first thing we want to do is read the header line, this will tell
+      // us the names of the fields, and how many input fields that there are.
+      String headLine = reader.readLine();
+      // Check to ensure that a header exists in our input file.
+      if (headLine == null) {
+        System.err.println("Invalid file format: no header provided");
+        closeStream(outStream);
+        return;
+      }
+      String[] header = parseFields(headLine, maxFields);
```
jmchristen, line 32: `header` is never used except to set `maxFields`—but you already know how many fields to expect, since you're only accepting one input format right now. I think you should ignore the header line and document on this function the expected sequence of fields in the input rows.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,80 @@
...
+        // Write the pose data to the necessary output stream.
+        currentPose.writeDelimitedTo(outStream);
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    closeStream(outStream);
+  }
+
+  private static void closeStream(OutputStream outStream) {
```
jmchristen, line 48: FYI, `OutputStream.close` is documented to do nothing.

https://docs.oracle.com/javase/8/docs/api/java/io/OutputStream.html#close--

src/main/java/proto/Pose.proto
```diff
@@ -0,0 +1,20 @@
...
+
+package formats;
+option java_package = "proto";
+option java_outer_classname = "SensorData";
+
+message Pose{
+  // Position is in WGS84 latitude, longitude, altitude.
+  double gps_timestamp = 1; // Time in seconds.
+  double lat = 2;  // Degrees in [-90, 90].
+  double lng = 3;  // Degrees in [-180,180].
```
jmchristen, line 15: Why allow multiple representations for the same location? Usually we would have these as half-open intervals.

src/main/java/proto/Pose.proto
```diff
@@ -0,0 +1,20 @@
...
+  double alt = 4; // Meters
+  float roll_deg = 5; // Degrees per second
+  float pitch_deg = 6; // Degrees per second
+  float yaw_deg = 7; // Degrees per second
```
jmchristen, line 19: I don't think these are supposed to be angular speeds.

src/main/java/proto/Pose.proto
```diff
@@ -0,0 +1,20 @@
...
+package formats;
+option java_package = "proto";
+option java_outer_classname = "SensorData";
+
+message Pose{
+  // Position is in WGS84 latitude, longitude, altitude.
+  double gps_timestamp = 1; // Time in seconds.
+  double lat = 2;  // Degrees in [-90, 90].
+  double lng = 3;  // Degrees in [-180,180].
+  double alt = 4; // Meters
```
jmchristen, line 16: "Meters above WGS84"? (Or a different specification, if it's not WGS84.)

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,112 @@
...
+    // decode.
+    // First encode data from text to proto format.
+    Encoder.encode(encoderInput, encoderOutput);
+
+    // Now decode the data that was just created in that file.
+    decoderInput = new FileInputStream(decoderInputFile);
+    Decoder.decode(decoderInput, decoderOutput);
+
+    // Open the file and create a String from its contents.
+    File verifyFile = new File(decoderOutputFile);
```
jmchristen, line 66: Interacting with the file system makes tests slower and flakier, so it's better to avoid it unless the actual behavior you're testing involves the file system. I think you can make these tests work without ever writing out to disk.

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,112 @@
...
+    testInputArray =
+        new String[] {
+          "gps_timestamp lat lng alt roll_deg pitch_deg yaw_deg",
+          "1140885738.407125 48.129872 11.582905 10.015034 -177.971817 -3.228588 10.700191",
+          "1140885738.475212 48.129872 11.582905 10.014367 -177.982559 -3.510518 10.743480"
+        };
+    testInput = testInputArray[0] + "\n" + testInputArray[1] + "\n" + testInputArray[2];
+    encoderInput = new ByteArrayInputStream(testInput.getBytes());
+    encoderOutput = new FileOutputStream(decoderInputFile);
+    decoderOutput = new FileOutputStream(decoderOutputFile);
```
jmchristen, line 47: Prefer to make these local variables in each test. It's much easier to visually verify a test when you can see the input and expected output right next to each other. (This is of course balanced against efficiency, but I don't think that's an issue here.)

https://docs.google.com/document/d/1vLre-K3xneE4ROU6Hwkm4r0ujJ9_uFkG01Zu4zaoRzw/edit?usp=sharing

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,112 @@
...
+    File out = new File(decoderOutputFile);
+    in.delete();
+    out.delete();
+  }
+
+  private void verifyLine(JsonObject readInLine, String[] expectedLine) {
+    // The amount to allow for rounding when Double conversions are taking place in
+    // the JUnit assertEquals for Double documentation.
+    Double epsilon = .1;
+    Assert.assertEquals(
```
jmchristen, line 97: This is one of the rare cases where we prefer to have a static import:
```
import static org.junit.Assert.assertEquals;
```
so that you can just write `assertEquals(a, b)` in your tests.

src/test/java/EncodeTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import proto.SensorData.Pose;
+
+@RunWith(JUnit4.class)
+/** Validate that a .txt file can be properly encoded into a .proto file. */
+public final class EncodeTest {
```
jmchristen, line 19: Same comments here as in `DecodeTest`.

src/main/java/proto/Pose.proto
D-moe: I was a bit confused on the units myself, Jie mentioned some of the units in a previous comment, but I don't think I completely understood what they were all suppose to be?

src/main/java/proto/Pose.proto
D-moe: I copied these degree values from a previous comment by Jie I'm not totally sure the reasoning behind this interval.

src/main/java/proto/Pose.proto
D-moe: Whoops

src/main/java/proto/Pose.proto
zjwoody: Yes it is "Meters over WGS84 datum"

src/main/java/proto/Pose.proto
zjwoody: This should be degrees.

src/main/java/proto/Pose.proto
```diff
@@ -9,12 +9,12 @@ option java_package = "proto";
 option java_outer_classname = "SensorData";
 
 message Pose{
-  // Position is in WGS84 latitude, longitude, altitude.
+  // Position is in Meters above WGS84 latitude, longitude, altitude.
```
jmchristen, line 12: Only altitude is meters above the WGS84 datum.

src/test/java/DecodeTest.java
```diff
@@ -1,75 +1,53 @@
...
-    decoderOutput = new FileOutputStream(decoderOutputFile);
-  }
-
-  @Test
-  /**
-   * Convert the encoded .sstable file (proto format) into json. Basic sanity checks, up to the user
-   * to make sure the output is indeed as expected.
-   */
-  public void validateDecoding() throws IOException, OutOfMemoryError, SecurityException {
+    String testInput = testInputArray[0] + "\n" + testInputArray[1] + "\n" + testInputArray[2];
```
jmchristen, line 33: `String.join` can do this for you.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,79 @@
...
+public class Encoder {
+  /*
+   * Convert a given .txt that contains a header line specifying file contents
+   * into the defined protobuf format.
+   */
+  public static void encode(InputStream inStream, OutputStream outStream) {
+    // Set an upper threshold by default on the maximum number of fields, this
+    // will make the call to split faster.
+    // The expected number of fields in our pose format.
+    final int maxFields = 7;
```
jmchristen, line 20: This can be a `static final` class member.

src/main/java/encoder/Encoder.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static Pose parsePose(String[] input) {
+    Pose.Builder pose = Pose.newBuilder();
+    pose.setGpsTimestamp(Double.parseDouble(input[0]))
+        .setLat(Double.parseDouble(input[1]))
+        .setLng(Double.parseDouble(input[2]))
+        .setAlt(Double.parseDouble(input[3]))
+        .setRollDeg(Float.parseFloat(input[4]))
+        .setPitchDeg(Float.parseFloat(input[5]))
+        .setYawDeg(Float.parseFloat(input[6]));
```
jmchristen, line 76: Please document the expected txt format in the class comment.

src/test/java/DecodeTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Validate that decoding the protobuf file to json works properly. */
+@RunWith(JUnit4.class)
+public final class DecodeTest {
+
+  @Test
+  /**
+   * Convert the encoded .sstable file (proto format) into json. Basic sanity checks, up to the user
```
jmchristen, line 22: There are some outdated references to files in the comments in this test and the encoder test.

zjwoody:
Hi David, could you merge this pr? It's already approve for few days.

PR #14, opened 2020-07-30
Draft datastorage API

freliciatucker:
- please Disregard  FileUploard Servlet and index.html
- attempts to create API for David to use to encode and decode blobs

### DIFF
--- a/src/main/java/com/google/cloud/storage/Database.java
+++ b/src/main/java/com/google/cloud/storage/Database.java
@@ -1,0 +1,31 @@
+/* This class creates an instance of a database that a file can be written to and read from. */
+public class Database{
+
+    /* Name of Database. */
+    private String name;
+
+    /* List of all buckets in Database. */
+    private Arraylist<Bucket> buckets; 
+    
+     /* Before we can use Google Cloud storage, we have to create a service object. */
+    private Storage storage;
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/storage/Database.java
```diff
@@ -0,0 +1,46 @@
+package com.google.cloud.storage;
+
+import java.util.HashMap;
+
+
+/* This class creates an instance of a database that a file can be written to and read from.
```
zjwoody, line 6: Use block comments:
https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

zjwoody, line 6: It is not creates an instance, it is the database api

src/main/java/com/google/cloud/storage/Database.java
```diff
@@ -0,0 +1,46 @@
...
+Assumptions: for weach runID, only 1 of each datatype can be associated with it */
+public class Database{
```
zjwoody, line 8: I think you may need to update the code by our yesterday's discussion.

freliciatucker: will be meeting today with David to discuss changes!

PR #15, opened 2020-08-05
Database, Reader, and Writer

freliciatucker:
- Database makes input and output streams which reader and writer call respectively. 
- still developing testing
- looking for suggestions for best data structure to hold list of files in database or even if this is needed


src/main/java/IO/DbReader.java
```diff
@@ -0,0 +1,53 @@
...
+    /* Comments out extension variable bc may not need it
+     private static final String extension = ".bin"; . */
+    private String fileName;
+    
+    /* Makes a reader object. */
+    public DbReader(Database database, String runId, String type) {
+        this.database = database;
+        this.runId = runId;
+        this.type = type;
+        fileName = database.findName(this.runId, this.type);
```
zjwoody, line 28: You don't need the fileName, since it's the implementation detail of db, the user doesn't need to know it, let db provide api to readData(runId, type)

src/main/java/IO/DbReader.java
```diff
@@ -0,0 +1,53 @@
...
+/* DbReader class accesses blob information to read its contents. */
+public class DbReader {
+
+    private Database database;
+    private String runId;
+    private String type;
+    private InputStream in;
+
+    /* Comments out extension variable bc may not need it
+     private static final String extension = ".bin"; . */
```
zjwoody, line 20: Wrong multiline comments style.

pom.xml
```diff
@@ -1,60 +1,59 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
zjwoody, line 1: Seems each line have extra blanks at the end.

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.Charset;
+import java.util.Random;
+
+import data.Database;
+
+/* DbWriter class */
```
zjwoody, line 13: Try to add more descriptions.

src/main/java/IO/DbReader.java
```diff
@@ -0,0 +1,53 @@
...
+    public DbReader(Database database, String runId, String type) {
+        this.database = database;
+        this.runId = runId;
+        this.type = type;
+        fileName = database.findName(this.runId, this.type);
+    }
+
+    /*
+     * Return a stream object for reading from the file, eventually from the
+     * database, right now reading a file that includes the database name.
```
zjwoody, line 33: Returns a stream object for reading from database.

src/main/java/IO/DbReader.java
```diff
@@ -0,0 +1,53 @@
...
+        return in;
+    }
+
+    /*
+     * Handle all actions that are necessary when done using the DbReader
+     * Object.
+     */
+    public void finish() {
+        try {
+            in.close();
```
zjwoody, line 46: What if the user close the stream themselves?

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+  private String fileName;
+  private static final int hashLength = 10;
+  private OutputStream out;
+  // private static final String extension = ".bin";
+
+  /* Creates instance of a Writer with this runId. */
+  public DbWriter(Database database, String runId, String type) {
+    this.database = database;
+    this.runId = generateRandomRunId();
+    fileName = database.newDatabaseEntry(this.runId, this.type);
```
zjwoody, line 28: Ditto, you don't need the filename.

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+
+  // Create writer object, then generate random runId if need one
+  // Method so that the user can get the value of the runId if they need one
+  public String getRunId() {
+    return runId;
+  }
+
+  /* Gets an outputstream from the database. */
+  public OutputStream write() {
+    out = database.writeData(fileName);
```
zjwoody, line 47: database.writeData(runId, type).

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+    return out;
+  }
+   
+   /* Returns unique runid. */
+  private String generateRandomRunId() {
```
zjwoody, line 52: Should let user specify the run_id.

src/main/java/data/Database.java
```diff
@@ -0,0 +1,74 @@
...
+    /* Place where files are stored.*/
+    private String name;
+
+    /* List of all files stored in Database.*/
+    ArrayList<String> files;
+ 
+    /* Creates instance of a database with this name. */
+    public Database(String name){
+      /* Database will be GCS and buckets. Bucket will named by different runids. */
+      this.name = name;
```
zjwoody, line 26: You may also need to init files by list all the files under the directory.

src/main/java/data/Database.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+
+/* This class creates an instance of a database that a file can be written to and read from.
+Assumptions: for each runID, only 1 of each datatype can be associated with it */
+public class Database{
+ 
+    /* Place where files are stored.*/
+    private String name;
```
zjwoody, line 18: name => baseDir
which is the directory that used to store all the data.

src/main/java/data/Database.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /* List of all files stored in Database.*/
+    ArrayList<String> files;
+ 
+    /* Creates instance of a database with this name. */
+    public Database(String name){
+      /* Database will be GCS and buckets. Bucket will named by different runids. */
```
zjwoody, line 25: The current database is the FileDatabase, if you want to add more other kinds of db, you can define new class e.g. GCSDatabase, then consider some abstract design.

testInput2.txt
```diff
@@ -0,0 +1,4 @@
+gps_timestamp lat lng alt roll_deg pitch_deg yaw_deg
```
zjwoody, line 1: Seems this file is unrelated to your current code, you can delete it.

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+/* DbWriter class */
+public class DbWriter {
+
+  private String runId;
+  private String type;
+  private Database database;
+  private String fileName;
+  private static final int hashLength = 10;
+  private OutputStream out;
+  // private static final String extension = ".bin";
```
zjwoody, line 22: the .bin usually stands for executable binary, you can use another extension e.g. recordio, or not use any extension.

testInput2.txt
freliciatucker: i planning on testing by write and reading this file to database.

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+    try {
+      out.close();
+    } catch (IOException e) {
+      System.err.println("Unable to close the file, an error occurred");
+      e.printStackTrace();
+    }
+  }
+
+  // Code from online to generate a random hash of length n.
+  private static String getAlphaNumericString(int n) {
```
jmchristen, line 84: Check out https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#randomUUID--

Also, use extreme caution when copying code from online. We always need to inspect the copyright first to make sure it's okay.

src/main/java/IO/DbWriter.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /* Creates instance of a Writer that doesnt have a runId yet. */
+  public DbWriter(Database database, String type) {
+    this.database = database;
+    this.type = type;
+    runId = generateRandomRunId();
+    fileName = database.newDatabaseEntry(this.runId, this.type);
+  }
+
+  // Create writer object, then generate random runId if need one
```
jmchristen, line 39: I think this comment is out of place.

src/main/java/com/google/cloud/storage/DatabaseQuery.java
```diff
@@ -0,0 +1,11 @@
+package google.cloud.storage;
+
+import data.Database;
+import java.util.ArrayList;
+
+public class DatabaseQuery{
```
jmchristen, line 6: What's the purpose of this class?

src/main/java/data/Database.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /* Returns projectID of Database. */
+  public String getName() {
+    return name;
+  }
+
+  // Attempt to open a stream to the data, normally to database, in this case to
+  // file.
+  public InputStream readData(String fileName) {
```
jmchristen, line 32: Document when a function's return parameter (or argument) may be null. Otherwise, we generally assume arguments and returns are non-nullable, because it's easier to reason about.

https://drive.google.com/file/d/1It0D63aBJZE_NGqMHgBTEZbtF0Y79WYI/view?usp=sharing

src/main/java/IO/DbWriter.java
D-moe: I'll keep this in mind, I was looking for a quick way to generate the key but will be more careful next time.

src/main/java/com/google/cloud/storage/DatabaseQuery.java
D-moe: The idea with this class is it handles general queries about runs for a specific database. I.e. all the runs that have been uploaded, their associated types, etc. While getting specific files the database instance is passed to the reader writer class. When getting information about the database entries themselves (i.e. likely through the spanner tables) the database instance is passed to the database query class.

src/main/java/IO/DbReader.java
freliciatucker: I'm thinking of adding to the comment that when using this API, user must use the finish() to close instead of closing stream directly, incase things need to be wrapped up when the user is done using the reader class this could all happen in finish

src/main/java/data/Database.java
freliciatucker: Will this be ok to do for the iteration after the MVP ?

src/main/java/com/google/cloud/storage/DatabaseQuery.java
freliciatucker: wanted to abstract away from the database class.

src/main/java/IO/DbWriter.java
freliciatucker: i was under the impression that the user would could upload a file with a runid, or without, so we wanted to make sure the user has both options

src/main/java/IO/DbReader.java
```diff
@@ -1,53 +1,54 @@
...
 
-import data.Database;
-
-/* DbReader class accesses blob information to read its contents. */
+/*
+ * DbReader class accesses blob information to read its contents with an
+ * InputStream. If user makes DbReader object and calls the read() function,
+ * user also must call close on the Inputstream returned. Instead, the user
+ * must use the finish() function to handle closing the inputstream and any
+ * other necessary end operations.
```
jmchristen, line 12: I don't quite understand this comment. Is the user supposed to call `InputStream.close()`, `DbReader.finish()`, or both?

src/main/java/IO/DbWriter.java
```diff
@@ -1,16 +1,15 @@
...
-import java.util.Random;
-
-import data.Database;
 
-/* DbWriter class */
+/* DbWriter class provides output stream from database to client.
+ * If user makes DbWriter object and calls the write() function,
+ * user also must call close on the Outputstream returned. Instead, the user
+ * must use the finish() function to handle closing the inputstream and any
+ * other necessary end operations.  */
```
jmchristen, line 12: Same question here as on `DbReader`.

src/main/java/IO/DbReader.java
freliciatucker: just DbReader. ill make more clear

src/main/java/IO/DbWriter.java
freliciatucker: same here

src/main/java/data/Database.java
```diff
@@ -0,0 +1,79 @@
...
+  /* Creates instance of a database with this name. */
+  public Database(String name) {
+    this.directory = name;
+    files = new ArrayList<String>();
+  }
+
+  /* Returns name associated with Database. */
+  public String getDatabase() {
+    return directory;
+  }
```
zjwoody, line 28: Who will use it? This api looks weird.

src/main/java/data/Database.java
```diff
@@ -0,0 +1,79 @@
...
+  /* Place where files are stored.*/
+  private String directory;
+
+  /* List of all files stored in Database.*/
+  private ArrayList<String> files;
+
+  /* Creates instance of a database with this name. */
+  public Database(String name) {
+    this.directory = name;
+    files = new ArrayList<String>();
```
zjwoody, line 22: I think it is better to list the files under the directory than init with empty list. Since if you restart your db service, you'll lost all existing files.

src/main/java/data/Database.java
```diff
@@ -0,0 +1,79 @@
...
+Assumptions: for each runID, only 1 of each datatype can be associated with it. */
+public class Database {
+  /* Place where files are stored.*/
+  private String directory;
+
+  /* List of all files stored in Database.*/
+  private ArrayList<String> files;
+
+  /* Creates instance of a database with this name. */
+  public Database(String name) {
```
zjwoody, line 20: name => basePath or baseDir
name looks not match with the actual usage.

freliciatucker:
I resolve all comments posted here on another branch. would it be possible to merge this branch to master, make a new branch with these commits and merge my other changes to this new branch in order to create smaller PRs that are easier to merge later? If possible, please approve this branch

jmchristen (dismissed):

PR #16, opened 2020-08-07
Servlets

D-moe:
Servlets for getting, uploading, etc. This was extracted from the MVP branch for careful review and pull request.

This will never pass the maven test as it depends on the database, etc, this is ok tho as it will all be merged in together.

### DIFF
--- a/src/main/java/servlets/DataEntries.java
+++ b/src/main/java/servlets/DataEntries.java
@@ -21,6 +21,8 @@ import shared.sharedObjects;
 public class DataEntries extends HttpServlet {
   // Return numEntries runId and name pairs for getting urls to load viewer data.
   // Optional numEntries parameter limits the number of returned values.
+  private Database dataInstance = new Database(sharedObjects.databaseName);
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("text/html;");
@@ -49,10 +51,8 @@ public class DataEntries extends HttpServlet {
     // could be recovered from the shared instance, this implementation will be
     // changed when merged into Frelica's database implementation and this function
     // will no longer be called.
-    Database dataInstance = new Database(sharedObjects.databaseName);
     String fileName = dataInstance.getName() + "/";
     // Remove extraneous whitespace from the name.
-    fileName.trim();
     File dir = new File(fileName);
     System.out.println(dir.getAbsolutePath());
     File[] dataEntries = dir.listFiles();
@@ -77,9 +77,9 @@ public class DataEntries extends HttpServlet {
       String type = params[1];
 
       // Add empty TreeSet to map if none is present.
-      idToDataTypes.computeIfAbsent(runId, id -> new TreeSet<String>());
+      TreeSet<String> typeList = idToDataTypes.computeIfAbsent(runId, id -> new TreeSet<String>());
       // Add specific entry.
-      idToDataTypes.get(runId).add(type);
+      typeList.add(type);
     }
     System.out.println(idToDataTypes);
     Gson gson = new GsonBuilder().setPrettyPrinting().create();
--- a/src/main/java/servlets/DataEntries.java
+++ b/src/main/java/servlets/DataEntries.java
@@ -2,6 +2,7 @@ package servlets;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import data.Database;
 import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
@@ -44,9 +45,12 @@ public class DataEntries extends HttpServlet {
   // future.
   private String getJson() {
     // In the future this will be implemented by calling a method from the database
-    // class. For now this is implemented as reading from a file.
-
-    String fileName = sharedObjects.dataInstance.getName() + "/";
+    // class. For now this is implemented as reading from a file. Although the name
+    // could be recovered from the shared instance, this implementation will be
+    // changed when merged into Frelica's database implementation and this function
+    // will no longer be called.
+    Database dataInstance = new Database(sharedObjects.databaseName);
+    String fileName = dataInstance.getName() + "/";
     // Remove extraneous whitespace from the name.
     fileName.trim();
     File dir = new File(fileName);
@@ -59,7 +63,7 @@ public class DataEntries extends HttpServlet {
 
     // Associate a runID with a list (for json) of the entry types that are found
     // for it.
-    HashMap<String, TreeSet<String>> dataMap = new HashMap<String, TreeSet<String>>();
+    HashMap<String, TreeSet<String>> idToDataTypes = new HashMap<String, TreeSet<String>>();
     for (File file : dataEntries) {
       // Parse through all entries and write them as JSON.
       String name = file.getName();
@@ -71,17 +75,15 @@ public class DataEntries extends HttpServlet {
       }
       String runId = params[0];
       String type = params[1];
-      if (!dataMap.containsKey(runId)) {
-        TreeSet<String> toAdd = new TreeSet<String>();
-        toAdd.add(runId);
-        dataMap.put(runId, toAdd);
-      } else {
-        dataMap.get(runId).add(type);
-      }
+
+      // Add empty TreeSet to map if none is present.
+      idToDataTypes.computeIfAbsent(runId, id -> new TreeSet<String>());
+      // Add specific entry.
+      idToDataTypes.get(runId).add(type);
     }
-    System.out.println(dataMap);
+    System.out.println(idToDataTypes);
     Gson gson = new GsonBuilder().setPrettyPrinting().create();
-    String json = gson.toJson(dataMap);
+    String json = gson.toJson(idToDataTypes);
     System.out.println(json);
     return json;
   }
--- a/src/main/java/servlets/DataEntries.java
+++ b/src/main/java/servlets/DataEntries.java
@@ -1,7 +1,7 @@
 package servlets;
 
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
+import com.google.appengine.repackaged.com.google.gson.Gson;
+import com.google.appengine.repackaged.com.google.gson.GsonBuilder;
 import data.Database;
 import java.io.File;
 import java.io.IOException;
@@ -19,20 +19,20 @@ import shared.sharedObjects;
  */
 @WebServlet("/data")
 public class DataEntries extends HttpServlet {
-  // Return numEntries runId and name pairs for getting urls to load viewer data.
-  // Optional numEntries parameter limits the number of returned values.
   private Database dataInstance = new Database(sharedObjects.databaseName);
 
+  // Get a list of the data values that are currently in the database. Take an optional number of
+  // entries parameter.
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    response.setContentType("text/html;");
+    response.setContentType("text/html");
     String entriesUnparsed = request.getParameter("numEntries");
     // If the numEntries value is not set default to getting all the entries.
     int numEntries = -1;
     if (entriesUnparsed != null) {
       try {
         numEntries = Integer.parseInt(entriesUnparsed);
-      } catch (Exception e) {
+      } catch (NumberFormatException e) {
         System.err.println("Invalid value provided stick to default");
         numEntries = -1;
       }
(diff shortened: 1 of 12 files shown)


src/main/java/servlets/Upload.java
```diff
@@ -0,0 +1,43 @@
...
+@WebServlet("/upload")
+public class Upload extends HttpServlet {
+  // Return numEntries runId and name pairs for getting urls to load viewer data.
+  // Optional numEntries parameter limits the number of returned values.
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException, IllegalStateException {
+    response.setContentType("text/html;");
+
+    String dataType = "pose";
```
zjwoody, line 27: You can get the data type from request.

src/main/webapp/index.html
```diff
@@ -1,11 +1,20 @@
...
-	<body>
-        <h1>Capstone project page</h1>
+	<body onload="loadUrls()">
+		<h1>Capstone project page</h1>
+		<form action="/upload" method="post" enctype="multipart/form-data">
+			Select image to upload:
+			<input type="file" name="file" id="fileToUpload">
+			<input type="submit" value="Upload Image" name="submit">
+		  </form>
+		  <ul id="myUL">
```
zjwoody, line 16: It is better to give a meaningful id name.

src/main/webapp/index.html
```diff
@@ -1,11 +1,20 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 	<head>
 		<meta charset="utf-8">
 		<title>My first three.js app</title>
```
zjwoody, line 5: May need to change the title to StellarPoint Project

src/main/webapp/index.html
```diff
@@ -1,11 +1,20 @@
...
+		<script src="script.js"></script>
 	</head>
-	<body>
-        <h1>Capstone project page</h1>
+	<body onload="loadUrls()">
+		<h1>Capstone project page</h1>
+		<form action="/upload" method="post" enctype="multipart/form-data">
+			Select image to upload:
+			<input type="file" name="file" id="fileToUpload">
+			<input type="submit" value="Upload Image" name="submit">
```
zjwoody, line 14: Maybe change the "image" to "run data"

src/main/java/shared/sharedObjects.java
```diff
@@ -0,0 +1,10 @@
+package shared;
+
+import data.Database;
+
+// Variables that need to be accessed between multiple class instances (i.e. Database initialization
+// variable).
+public class sharedObjects {
+  // public static
+  public static Database dataInstance = new Database("testDB");
```
jmchristen, line 9: Is there a reason this needs to be shared between classes? It's not unusual to have multiple classes with their own `Database` objects referring to the same underlying database, and it's generally simpler.

src/main/java/servlets/DataEntries.java
```diff
@@ -0,0 +1,90 @@
...
+      String hash = params[0];
+      String type = params[1];
+      if (!dataMap.containsKey(hash)) {
+        TreeSet<String> toAdd = new TreeSet<String>();
+        toAdd.add(type);
+        dataMap.put(hash, toAdd);
+      } else {
+        dataMap.get(hash).add(type);
+      }
+    }
```
jmchristen, line 83: This block needs some commentary and more descriptive names (factoring it out into a function might help). E.g. what does `dataMap` represent? What are `hash` and `type`? Why do the entries need to be ordered?

src/main/java/servlets/RunInfo.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import shared.sharedObjects;
+
+/** Fetch the given data points for a specific run based on its runId and the dataType. */
+@WebServlet("/getrun")
+public class RunInfo extends HttpServlet {
+  // Return numEntries runId and name pairs for getting urls to load viewer data.
+  // Optional numEntries parameter limits the number of returned values.
```
jmchristen, line 16: Erroneously copied comment?

src/main/java/servlets/Upload.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.Part;
+import shared.sharedObjects;
+
+/** Fetch the given data points for a specific run based on its runId and the dataType. */
+@MultipartConfig
+@WebServlet("/upload")
+public class Upload extends HttpServlet {
+  // Return numEntries runId and name pairs for getting urls to load viewer data.
+  // Optional numEntries parameter limits the number of returned values.
```
jmchristen, line 21: Erroneously copied comment?

src/main/java/servlets/Upload.java
D-moe: Yeah totally agree, for mvp I think we are assuming just pose data, but I'll change this when we handle other forms.

src/main/java/shared/sharedObjects.java
D-moe: My reasoning here was that it's inefficient to construct a new database object, might as well share the object as we don't know how costly setting up the initial connection may be.

D-moe: Though I guess I would have thought this technique was simpler than having two places where you have to refer to the same database explicitly by name.

jmchristen: My biggest concern is thread safety (though I'm not sure whether that's actually an issue in this particular case). We can expect multiple `Database` objects to coordinate their work behind the scenes, but if you _also_ want the `Database` object itself to be threadsafe, you've created a lot more work for yourself.

Re: inefficiency, I'd consider it premature optimization. The number of DB connections scales with number of server instances regardless, so you're only changing things by a constant factor. If the app starts up slowly with the naive implementation, I'd profile it to make sure that's actually where the bottleneck is.

Re: simplicity, it may indeed be simpler to have a single source of truth for the database name. But that should probably be a string constant, or a string in a config file, not necessarily an object. And I think it's simpler to reason about lifetimes when you keep as much stuff as possible local or explicitly passed in to your classes.

If it _is_ worthwhile to share a `Database` object, we would generally construct it in some initialization function that can pass it as a constructor parameter to the classes that need it. That tends to make testing easier. (But I'm not sure offhand how it would work with this framework.)

src/main/java/servlets/DataEntries.java
```diff
@@ -61,24 +57,26 @@ private String getJson() {
...
-      if (!dataMap.containsKey(hash)) {
+      if (!dataMap.containsKey(runId)) {
         TreeSet<String> toAdd = new TreeSet<String>();
-        toAdd.add(type);
-        dataMap.put(hash, toAdd);
+        toAdd.add(runId);
+        dataMap.put(runId, toAdd);
       } else {
-        dataMap.get(hash).add(type);
+        dataMap.get(runId).add(type);
```
jmchristen: I think you could simplify this with `HashMap.computeIfAbsent`.

```
dataMap.computeIfAbsent(runId, id -> new TreeSet<String>()).add(type);
```

src/main/java/servlets/DataEntries.java
```diff
@@ -61,24 +57,26 @@ private String getJson() {
       return "{}";
     }
 
+    // Associate a runID with a list (for json) of the entry types that are found
+    // for it.
     HashMap<String, TreeSet<String>> dataMap = new HashMap<String, TreeSet<String>>();
```
jmchristen, line 62: `dataMap` needs a more descriptive name. Maybe `idToDataTypes`?

jmchristen, line 62: Does `TreeSet` get converted to a JSON array, or something more complicated?

src/main/java/shared/sharedObjects.java
D-moe: The multithreaded point makes a lot of sense thanks for pointing that out I tend to overlook those concerns so I'm glad you brought it to my attention so I can be more aware of these issues in the future. I'll implement this change.

src/main/java/servlets/DataEntries.java
D-moe: Yeah simple JSON array is the conversion GSON performs

src/main/java/servlets/DataEntries.java
D-moe: > Can you put the new homepage stuff in a separate PR? It's not closely related to the servlets, and reviews will go quicker if the PR doesn't keep growing.

Some of the initial comments on the pr were regarding features in the home page, naming conventions etc, which is why I committed these features into this branch. Should I not address those comments then? I definitely think it makes sense to have it a separate pr, but also it doesn't make sense to change things I've already written while working on this branch?

jmchristen: Ah, I didn't remember that. It's good to address comments, but try not to add new features to PRs in review. I think "Add searchable feature" is what added most of the new code since we started review.

src/main/java/shared/sharedObjects.java
D-moe: Question about implementation of java servlets, is it true that the database instances scale with connections, or would one static instance be used through all instances? I assume its the former but I'm not actually certain about that?

D-moe: either way I'm swapping the implementation over

jmchristen: I'm not sure I understand your question. My assertion above was that each server (not servlet) instance must connect to the DB at least once. I think that's true regardless of the servlet framework, because the servers are isolated from each other.

D-moe: I guess I was thinking it may be possible for each client connecting to our server to share the same database connection. My question was mainly in regard to whether a static variable defined the way I did was created for each server instance, or only the servlet itself and its variables are created for each connection.

src/main/webapp/script.js
```diff
@@ -0,0 +1,87 @@
...
+    // Grab the runId entry.
+    td = tr[i].getElementsByTagName('td')[0];
+    if (td) {
+      txtValue = td.textContent || td.innerText;
+      console.log('The text value is ' + txtValue);
+      if (txtValue.toUpperCase().startsWith(filter)) {
+        tr[i].className = 'visible';
+        if (even) {
+          tr[i].className += ' even';
+        }
```
jmchristen, line 80: I think you can use `nth-child` to do this entirely in CSS.

https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

D-moe: I originally did this, the issue is that upon filtering the elements there is no way to refresh the style of those elements, so you get a random striped pattern.

src/main/webapp/script.js
```diff
@@ -0,0 +1,87 @@
...
+    const columnOne = document.createElement('th');
+    columnOne.innerText = 'RunID';
+    const columnTwo = document.createElement('th');
+    columnTwo.innerText = 'DataType';
+    headerRow.appendChild(columnOne);
+    headerRow.appendChild(columnTwo);
+    table.appendChild(headerRow);
+    let even = false;
+    for (const key in json) {
+      if (json.hasOwnProperty(key)) {
```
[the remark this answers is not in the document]
D-moe: The linter requires this check, originally I did not have it.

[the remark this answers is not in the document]
D-moe: According to their documentation there is an edge case where this syntax can cause error so you should check it from what I could gather, though I was a bit confused on this.

src/main/java/servlets/DataEntries.java
```diff
@@ -0,0 +1,90 @@
...
+  private String getJson() {
+    // In the future this will be implemented by calling a method from the database
+    // class. For now this is implemented as reading from a file. Although the name
+    // could be recovered from the shared instance, this implementation will be
+    // changed when merged into Frelica's database implementation and this function
+    // will no longer be called.
+    Database dataInstance = new Database(sharedObjects.databaseName);
+    String fileName = dataInstance.getName() + "/";
+    // Remove extraneous whitespace from the name.
+    fileName.trim();
```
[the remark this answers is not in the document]
D-moe: The trim probably isn't necessary, I'll remove it. Leftover from debugging.

D-moe:
could you approve these changes Jie so Frelicia can get her changes merged on top of mine?

PR #18, opened 2020-08-10
forked from Reader Writer and Database branch

freliciatucker:
- updates API for Reader Writer and Database as well as adds a test file
- responds to comments on put on Reader Writer Database branches ie. using Files API to keep track of files in a directory instead of attempting to do so on my own


src/main/java/data/Database.java
```diff
@@ -1,30 +1,35 @@
...
-    this.directory = name;
-    files = new ArrayList<String>();
+    folder = new File(name);
+    folder.mkdir();
   }
 
   /* Returns name associated with Database. */
-  public String getDatabase() {
-    return directory;
+  public File getDatabase() {
```
freliciatucker, line 26: this method is functionally redundant to the getDirectoryName() method. will delete

src/main/java/data/Database.java
```diff
@@ -55,25 +60,25 @@ public OutputStream writeData(String runId, String type) {
...
-    return files.get(0);
+    return makeFileName(runId, type);
   }
 
   /* Returns list of files in database. */
-  public ArrayList<String> getAllFiles() {
-    return files;
+  public List<File> getAllFiles() {
+    File[] files = getDatabase().listFiles();
+    return Arrays.asList(files);
```
freliciatucker, line 82: instead of keeping and updating my own list of files added to the directory, i call Arrays.asList, which is more accurate and efficient than my original method.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,66 @@
...
+  @Test
+  public void validateWriterRunId() {
+    System.out.println("validateWriterRunId()");
+    /* Validate that writer's runId is test. */
+    Assert.assertEquals(writer.getRunId(), "test");
+  }
+
+  @Test
+  public void validateReaderWriterDatabaseConnect() throws IOException {
+    System.out.println("validateReaderWriterDatabaseConnect()");
```
freliciatucker, line 47: any suggestions for testing? this one had errors and i was unsure of how to solve them. I was hoping to store the string "This is a test string". into my database and read and write from it. I wanted to test if the stream input and output return was equal, but the code did not compile.

D-moe: What was the issue with compilation, that seems like a reasonable suggestion?

src/main/java/data/Database.java
```diff
@@ -55,25 +60,25 @@ public OutputStream writeData(String runId, String type) {
...
     String fileName = makeFileName(runId, type);
-    files.add(fileName);
     return fileName;
   }
 
   /* Return name of file with runid and type in database. */
   public String findName(String runId, String type) {
-    String fileName = getDatabase() + "/" + runId + "_" + type;
-    return files.get(0);
+    return makeFileName(runId, type);
```
D-moe, line 76: What's the point of findName isn't it redundant with makeFileName? Why not just make makeFileName public and call that directly?

src/main/java/data/Database.java
D-moe: What's the motivation for storing the files object, the database will have to return a list of all its entries as one of its methods, which makes sense to return as a string? I imagine there is additional overhead for storing all these file objects, could you explain what benefit there would be. If the conclusion if I want to open a file I already have the object I don't think that makes more sense than just creating the file object when you need it?

src/main/java/data/Database.java
freliciatucker: that makes sense, ill consolidate to file name method

freliciatucker:
@D-moe im a little confused with your last comment. Are you suggesting changing the return type from list<File> to list<String> ? 
and for storing file objects, i try not to do that by just calling the listfiles method in the files class instead of keep a list<File> private variable that would just take up space

src/test/java/ReaderWriterTest.java
freliciatucker: update: it compiles now! there was an issue with one of the DbWriter's constructor

D-moe:
All i was saying in my last comment is before you returned an arraylist of strings now you return a list of files. I was wondering why it was worth storing the file objects and not just the list of strings. Is there some benefit you were anticipating from this change?

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+/* Validate that a .txt file can be properly be stored written and read from database.  */
+public final class ReaderWriterTest {
```
batldre, line 25: Looks good to me, only suggestion is to add more comments specifically in this file and try to break down the whole process of writing and reading. Minor things like explaining why you included the setup method with nothing in it, or why you created a second reader and writer while hardly using the first. This would help me process the whole file a lot better.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+      output1.write(b);
+    }
+    writer2.finish();
+
+    String readTest = null;
+    try (InputStream input = reader2.read()) {
+      StringBuilder textBuilder = new StringBuilder();
+      try (Reader reader =
+          new BufferedReader(
+              new InputStreamReader(input, Charset.forName(StandardCharsets.UTF_8.name())))) {
```
jmchristen, line 72: If both of these `try`s are necessary (see above comment; I'm not sure that they are), I think you can combine them into one. See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Test
+  public void validateWriterRunId() {
+    System.out.println("validateWriterRunId()");
+    /* Validate that writer's runId is test. */
+    Assert.assertEquals(writer.getRunId(), "test");
+  }
+
+  @Test
+  public void validateReaderWriterDatabaseConnect() throws IOException {
```
jmchristen, line 54: I'm a little confused reading this test with all of the variables with names starting with `read` or `write` and the nested logic. Can you try to streamline and maybe add some comments on what's happening?

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+    reader2.finish();
+
+    Assert.assertEquals(writeTest, readTest);
+  }
+
+  @Test
+  public void validateDatabaseQueryAfterTestFileIsAdded() throws IOException {
+    List<File> files = DatabaseQuery.getAllFiles(database);
+
+    Assert.assertTrue(true);
```
jmchristen, line 89: This test isn't doing anything.

src/main/java/data/Database.java
```diff
@@ -55,25 +60,20 @@ public OutputStream writeData(String runId, String type) {
...
+  /*Returns appropiate file name for data with this runId and type. */
+  private String FileName(String runId, String type) {
+    return getDirectoryName() + "/" + runId + "_" + type;
   }
 
   /* Adds file to database given its runid and type. */
   public String newDatabaseEntry(String runId, String type) {
-    String fileName = makeFileName(runId, type);
-    files.add(fileName);
+    String fileName = FileName(runId, type);
```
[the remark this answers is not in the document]
freliciatucker: function should be deleted

src/main/java/data/Database.java
```diff
@@ -1,38 +1,43 @@
...
-    files = new ArrayList<String>();
+    folder = new File(name);
+    folder.mkdir();
   }
 
   /* Returns name associated with Database. */
-  public String getDatabase() {
-    return directory;
+  public File getDatabase() {
+    return folder;
```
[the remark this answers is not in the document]
freliciatucker: i made public here bc i was think about how it could be accessed once the database was abstracted when we switch from b/w a file system and a GCS, but i realize now there is a way to make a database interface and still had a file database class that has this method hidden in it

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Before
+  public void setUp() {}
+
+  @Test
+  public void validateDatabaseName() {
+    System.out.println("validateDatabaseName()");
+    /* Validate that place where data is stored is named test. */
+    System.out.println("test = " + database.getDatabase());
+    Assert.assertEquals(database.getDirectoryName(), "test");
```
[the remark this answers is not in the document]
freliciatucker: i see that this static import reduces the amount of writing needed by taking out the extra Assert word at the beginning, but does it have anymore benefits?

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+/* Validate that a .txt file can be properly be stored written and read from database.  */
+public final class ReaderWriterTest {
+  private static String inputFile;
+  private static String outputFile;
+  private static InputStream input;
+  private static OutputStream output;
+
+  private Database database = new Database("test");
+  private DbWriter writer = new DbWriter(database, "test", "writingIn");
+  private DbReader reader = new DbReader(database, "test", "writingIn");
```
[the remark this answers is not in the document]
freliciatucker: i deleted reader and made writer local, but kept database global bc its used in two tests . tried to be "damp" in the the 3rd test.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Test
+  public void validateReaderWriterDatabaseConnect() throws IOException {
+    System.out.println("validateReaderWriterDatabaseConnect()");
+    DbWriter writer2 = new DbWriter(database, "test", "writingIn");
+    DbReader reader2 = new DbReader(database, "test", "writingIn");
+
+    String writeTest = "This is a test string";
+    /*Validate that what writer wrote to dtabase is what reader reads. */
+    try (OutputStream output1 = writer2.write()) {
```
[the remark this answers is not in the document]
freliciatucker: i was saying the user is not suppose to close the stream by directly calling stream.close() and to instead to call writer.finish() which closes the stream. But i also talked to David and he said he directly closed the stream. So I'll go in to update the comment to say either call stream.close or writer.finish but do not call both

src/test/java/ReaderWriterTest.java
freliciatucker: you are correct. im trying add this test to check that the databasequery method returns all files within a database. will add a TODO comment

src/main/java/data/Database.java
```diff
@@ -55,25 +62,17 @@ public OutputStream writeData(String runId, String type) {
...
-    return files.get(0);
+  /*Returns appropiate file name for data with this runId and type. */
+  private String fileName(String runId, String type) {
+    return getDatabaseName() + "/" + runId + "_" + type;
   }
 
   /* Returns list of files in database. */
   public ArrayList<String> getAllFiles() {
-    return files;
+    File[] files = getDatabase().listFiles();
```
zjwoody, line 72: member function can use member variable directly:
files = folder.listFiles();

src/main/java/data/Database.java
```diff
@@ -55,25 +62,17 @@ public OutputStream writeData(String runId, String type) {
...
+    List<File> filesAsFiles = Arrays.asList(files);
```
zjwoody, line 73: I think you don't need to create another list, you can just use files
for (int i = 0; i < files.length; i++) {
   filesAsStrings.add(files[i].getName());
}

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+/* Validate that a .txt file can be properly be stored written and read from database.  */
+public final class ReaderWriterTest {
+
+  @Test
+  public void validateDatabaseName() throws IOException {
+    System.out.println("validateDatabaseName()");
+    Database database = new Database("test");
+    /* Validate that place where data is stored is named test. */
+    System.out.println("test = " + database.getDatabaseName());
+    Assert.assertEquals(database.getDatabaseName(), "test");
```
zjwoody, line 32: assertEquals(expected, actual)
The first one the expected value, the next one is your actual value.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Test
+  public void validateWriterRunId() throws IOException {
+    System.out.println("validateWriterRunId()");
+    Database database = new Database("test");
+    /* Validate that writer's runId is test. */
+    DbWriter writer = new DbWriter(database, "test", "writingIn");
+    assertEquals(writer.getRunId(), "test");
```
zjwoody, line 41: assertEquals(expected, actual)

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+    System.out.println("test = " + database.getDatabaseName());
+    Assert.assertEquals(database.getDatabaseName(), "test");
+  }
+
+  @Test
+  public void validateWriterRunId() throws IOException {
+    System.out.println("validateWriterRunId()");
+    Database database = new Database("test");
+    /* Validate that writer's runId is test. */
+    DbWriter writer = new DbWriter(database, "test", "writingIn");
```
zjwoody, line 40: Give an explicit name of runid and type, otherwise it will be confused with other parameters name.
e.g. new DbWriter(database, "run_id_a", "Pose");

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Test
+  public void validateReaderWriterDatabaseConnect() throws IOException {
+    System.out.println("validateReaderWriterDatabaseConnect()");
+
+    /*Validate that what writer wrote to database is what reader reads. */
+    /* Initialize Database, reader, and writer. */
+    Database database2 = new Database("test2");
+    DbWriter writer2 = new DbWriter(database2, "test", "writingIn");
```
zjwoody, line 51: Try to give a better parameter name.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    assertEquals(writeTest, readTest);
+  }
+
+  @Test
+  public void validateDatabaseQueryAfterTestFileIsAdded() throws IOException {
+    /* TODO: Test if DatabaseQuery method returns all files stored in the database.*/
+    Database database = new Database("test");
+    ArrayList<String> files = DatabaseQuery.getAllFiles(database);
+    Assert.assertTrue(true);
```
zjwoody, line 89: Try to add some files and check the file list.

src/test/java/ReaderWriterTest.java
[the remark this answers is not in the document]
jmchristen: That's all. It just removes some redundant boilerplate, since in the context of a test it's very clear what the bare `assertEquals` means.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,102 @@
...
+    /* Delete directory. */
+  }
+
+  @Test
+  public void validateDatabaseQuery() throws IOException {
+    /* TODO: Test if DatabaseQuery method returns all files stored in the database.*/
+    Database database = new Database("validateDatabaseQuery");
+
+    ArrayList<String> expectedFiles = new ArrayList<String>();
+    ArrayList<String> files = DatabaseQuery.getAllFiles(database);
```
zjwoody, line 96: Write some files into the db, and check the list.

src/test/java/ReaderWriterTest.java
```diff
@@ -1,91 +1,102 @@
...
       output1.write(b);
     }
+    /* Close outputStream. */
     writer2.finish();
 
     String readTest = null;
-    try (InputStream input = reader2.read()) {
+
+    /* Read contents of file written to database. */
+    try (Reader reader =
```
jmchristen: Same here, the stream will be automatically closed and you also call `reader2.finish()`.

src/test/java/ReaderWriterTest.java
```diff
@@ -1,91 +1,102 @@
...
+    /* TODO: Test if DatabaseQuery method returns all files stored in the database.*/
+    Database database = new Database("validateDatabaseQuery");
+
+    ArrayList<String> expectedFiles = new ArrayList<String>();
+    ArrayList<String> files = DatabaseQuery.getAllFiles(database);
+
+    assertEquals(expectedFiles, files);
 
-    Assert.assertTrue(true);
+    /* Delete directory.*/
```
jmchristen: Extraneous comment.

src/main/java/IO/DbWriter.java
```diff
@@ -7,9 +7,10 @@
 
 /* DbWriter class provides output stream from database to client.
  * If user makes DbWriter object and calls the write() function,
- * user SHOULD NOT close the Outputstream returned directly. Instead, the user
- * must use the finish() function to handle closing the inputstream and any
- * other necessary end operations.  */
+ * user either close the Outputstream returned directly or
+ * use the finish() function to handle closing the inputstream and any
+ * other necessary end operations, BUT DO NOT CALL BOTH.
```
jmchristen, line 12: This API will be tricky for a user to use correctly, because they have to find and read this comment, then make the choice without any clear guidelines on how to choose. Always try to streamline APIs to make them harder to misuse, and prefer not to make users obey comments (which they often will miss) except as a last resort.

In this particular case I don't think you need to do anything special at all. Both `InputStream` and `OutputStream` say their `close` method does nothing.

https://docs.oracle.com/javase/8/docs/api/java/io/OutputStream.html#close--
https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#close--

batldre:
Looks good!

src/main/java/IO/DbWriter.java
freliciatucker: oh wowwww, interesting. will ask more in pod sync

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,126 @@
...
+    writer1.write();
+    expectedFiles.add(writer1.getRunId() + "_" + writer1.getType());
+
+    DbWriter writer2 = new DbWriter(database, "2ndFile", "");
+    writer2.write();
+    expectedFiles.add(writer2.getRunId() + "_" + writer2.getType());
+
+    DbWriter writer3 = new DbWriter(database, "3rdFile", "");
+    writer3.write();
+    expectedFiles.add(writer3.getRunId() + "_" + writer3.getType());
```
zjwoody, line 112: You can try to write a for loop.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,126 @@
...
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+/* Validate that a .txt file can be properly be stored written and read from database.  */
```
zjwoody, line 24: Could you check all your comments format?
/* comment. /*
one blank between /* and comment and */

src/main/java/data/Database.java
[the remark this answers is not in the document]
jmchristen: This isn't resolved; I'm still not sure why the `File` is exposed to users of `Database`. It looks like the function is only used in tests to clean up. Maybe you can write a `delete` method on `Database` instead?

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /* Add files to Database. */
+    int numOfFiles = 3;
+    String tempRunId;
+    String tempType;
+    for (int i = 0; i < numOfFiles; i++) {
+      tempRunId = "FileNo." + i;
+      tempType = "test";
+      DbWriter writer1 = new DbWriter(database, tempRunId, tempType);
+      writer1.write();
```
jmchristen, line 108: What's the point of `writer1.write()` when you discard the return value?

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,125 @@
...
+    ArrayList<String> expectedFiles = new ArrayList<String>();
+
+    /* Add files to Database. */
+    int numOfFiles = 3;
+    String tempRunId;
+    String tempType;
+    for (int i = 0; i < numOfFiles; i++) {
+      tempRunId = "FileNo." + i;
+      tempType = "test";
+      DbWriter writer1 = new DbWriter(database, tempRunId, tempType);
```
jmchristen, line 107: This can just be named `writer`.

src/test/java/ReaderWriterTest.java
```diff
@@ -0,0 +1,125 @@
...
+      writer1.write();
+      expectedFiles.add(writer1.getRunId() + "_" + writer1.getType());
+    }
+
+    /* Checks to see if all files are in Database. */
+    ArrayList<String> files = DatabaseQuery.getAllFiles(database);
+    Assert.assertTrue(expectedFiles.containsAll(files));
```
jmchristen, line 114: You can `static import` for `assertTrue` as well.

src/test/java/ReaderWriterTest.java
freliciatucker: during write() a file is created with the filename corresponding to the runid and type. this test is only testing that the files are all there, so it doesnt need to know about the contents of them.

src/main/java/data/Database.java
```diff
@@ -60,20 +58,28 @@ public OutputStream writeData(String runId, String type) {
...
 
-  /* Adds file to database given its runid and type. */
-  public String newDatabaseEntry(String runId, String type) {
-    String fileName = FileName(runId, type);
-    return fileName;
+  /* Returns list of files in database. */
+  public ArrayList<String> getAllFiles() {
+    File[] files = folder.listFiles();
+    ArrayList<String> filesAsStrings = new ArrayList<String>();
+    for (int i = 0; i < files.length; i++) filesAsStrings.add(files[i].getName());
```
jmchristen, line 70: Prefer to use for-each loops by default, instead of manually looping over indices.

https://docs.oracle.com/javase/8/docs/technotes/guides/language/foreach.html

src/test/java/ReaderWriterTest.java
```diff
@@ -1,91 +1,112 @@
...
-  private Database database = new Database("test");
-  private DbWriter writer = new DbWriter(database, "test", "writingIn");
-  private DbReader reader = new DbReader(database, "test", "writingIn");
-
-  @Before
-  public void setUp() {}
 
   @Test
-  public void validateDatabaseName() {
+  public void validateDatabaseName() throws IOException {
```
jmchristen, line 27: In tests, you can just write `throws Exception`. No one is writing code to call the test methods, so you don't have to document precisely which exceptions can be thrown.

src/test/java/ReaderWriterTest.java
```diff
@@ -1,91 +1,112 @@
...
+    Database database = new Database("validateConnect");
+    DbWriter writer = new DbWriter(database, "3rdTestRunId", "writingIn");
+    DbReader reader = new DbReader(database, "3rdTestRunId", "writingIn");
 
     String writeTest = "This is a test string";
-    /*Validate that what writer wrote to dtabase is what reader reads. */
-    try (OutputStream output1 = writer2.write()) {
+
+    /* Write writeTest into database. */
+    try (OutputStream output1 = writer.write()) {
```
jmchristen, line 63: This can just be `output`.

src/main/java/data/DatabaseQuery.java
```diff
@@ -0,0 +1,62 @@
...
+    /*
+     * [data: {runId: aks;ldfja entries: []}]
+     */
+
+    String fileName = database.getDatabaseName() + "/";
+    fileName.trim();
+    /*System.out.println("The filename is " + fileName);
+    File dir = new File(fileName);
+    System.out.println(dir.getAbsolutePath());
+    */
```
jmchristen, line 29: Extraneous comment.

src/main/java/data/DatabaseQuery.java
```diff
@@ -0,0 +1,62 @@
...
+    ArrayList<String> filenames = getAllFiles(database);
+    if (filenames == null) {
+      System.out.println("Directory is empty");
+      return "{}";
+    }
```
jmchristen, line 34: Alternately, you could define `getAllFiles` to never return null. Returning an empty `ArrayList` when there are no files is sensible, and means that the caller doesn't have to do this null check.

src/main/java/data/DatabaseQuery.java
```diff
@@ -0,0 +1,62 @@
...
+
+    HashMap<String, TreeSet<String>> dataMap = new HashMap<String, TreeSet<String>>();
+    for (String file : filenames) {
+      // Parse through all entries and write them as JSON
+      System.out.println("The filename is " + file);
+      String[] params = file.split("[_]");
+      for (String param : params) System.out.println(param);
+      if (params.length != 2) {
+        System.err.println("Invalid database entry format!");
+        return "";
```
jmchristen, line 44: You return "{}" for the `filenames == null` error case above, but "" for this error case. Is the difference intentional?

src/main/java/data/FileDatabase.java
```diff
@@ -0,0 +1,43 @@
...
+/* This class creates an instance of a database that a file can be written to and read from.
+Assumptions: for each runID, only 1 of each datatype can be associated with it. */
+public class FileDatabase implements Database {
+  /* Folder for Database. */
+  File folder;
+
+  /* Creates instance of a database with this name. */
+  public FileDatabase(String name) throws IOException {
+    folder = new File(name);
+    folder.mkdir();
```
jmchristen, line 21: Check for failure here. `File.mkdir` can return false, which would mean that all the other operations on the database are invalid.

https://docs.oracle.com/javase/8/docs/api/java/io/File.html#mkdir--

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+  private Bucket bucket;
+
+  /* Primary bucketname.*/
+  private String bucketName;
+
+  /* Before we can use Google Cloud storage, we have to create a service object. */
+  private Storage storage;
+
+  /* ProjectID. */
+  private String projectId = "pose-3d-viewer-step-2020";
```
jmchristen, line 33: Can this be `private static final`?

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+   */
+  private final GcsService gcsService =
+      GcsServiceFactory.createGcsService(
+          new RetryParams.Builder()
+              .initialRetryDelayMillis(10)
+              .retryMaxAttempts(10)
+              .totalRetryPeriodMillis(15000)
+              .build());
+
+  //  private Credentials credentials;
```
jmchristen, line 57: Extraneous comment.

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+
+  /*
+   * See this documentation for other valid locations:
+   * http://g.co/cloud/storage/docs/bucket-locations#location-mr
+   */
+  private String location = "US";
```
jmchristen, line 63: `location` is unused.

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+
+  /* Used below to determine the size of chucks to read in. Should be > 1kb and < 10MB. */
+  private static final int BUFFER_SIZE = 2 * 1024 * 1024;
+
+  /* These two next variables are copied from GCS tutorials. */
+  /*
+   * See the StorageClass documentation for other valid storage classes:
+   * https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/storage/StorageClass.html
+   */
+  private StorageClass storageClass = StorageClass.STANDARD;
```
jmchristen, line 43: `storageClass` is unused.

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+    System.out.println("in constructor");
+
+    System.out.println("initializing storage");
+
+    storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    System.out.println("making bucket");
+    bucketName = name;
+    bucket = storage.get(bucketName);
+
+    if (bucket == null) bucket = storage.create(BucketInfo.of(bucketName));
```
jmchristen, line 76: Always use the multiline `if` style with braces.

https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /* uploads file with objectname to GCSDatabase. */
+  /*private OutputStream uploadFile(String objectName) throws IOException {
+    GcsFilename filename = new GcsFilename(bucketName, objectName);
+    GcsFileOptions instance = GcsFileOptions.getDefaultInstance();
+    GcsOutputChannel outputChannel;
+    outputChannel = gcsService.createOrReplace(filename, instance);
+    return Channels.newOutputStream(outputChannel);
+  }*/
```
jmchristen, line 105: Extraneous comment.

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+  private OutputStream uploadFile(String objectName) throws IOException {
+
+    Blob blob = bucket.create(objectName, new byte[0]);
+    WriteChannel writeChannel = blob.writer();
+    OutputStream out = Channels.newOutputStream(writeChannel);
+    System.out.println(out);
+    blob = blob.reload();
```
jmchristen, line 112: I don't think you need `blob.reload()` since `blob` is a local variable. (Also, I think the examples suggest to call `reload` after writing to the bucket, but in your case the write doesn't happen within this function.)

src/main/java/data/GCSDatabase.java
```diff
@@ -0,0 +1,165 @@
...
+         Blob objectCreated = bucket.create(objectInfo, topicMessagesMap.get(topicName).getBytes());
+         Blob blob = bucket.create(objectName, new byte[0]);
+         WriteChannel writeChannel = blob.writer();
+         OutputStream out = Channels.newOutputStream(writeChannel);
+         System.out.println(out);
+         out.write(5);
+         blob = blob.reload();
+         return out;
+    }
+  */
```
jmchristen, line 128: Extraneous comment.

src/main/java/data/FileDatabase.java
freliciatucker: thanks! i added a try catch to resolve. i  thought by adding the IOExeption of the line of the method signature that would check for failure. should i take this out now that i have the try catch?

jmchristen: (Note: I don't see any new commits since my review, so I can't see the change you're referring to.)

Adding `throws IOException` to the method signature only signifies that it's _allowed_ for the method to throw that exception. It's mostly a form of documentation for the callers, so that they know the method can fail. It doesn't actually change the behavior of the function at all, and it's possible to have a method marked `throws IOException` that is actually incapable of throwing. That's what you have here.

The compiler _does_ enforce the converse: if a method can throw an exception, the method signature must include the `throws` clause. Consider these two methods.
```
private int f() {
  return 0;
}
private int g() throws IOException {
  return 0;
}
```
It's not possible for either method to throw an exception, because nothing in either method body would cause it. The two methods behave identically. The difference is in how callers have to behave.
```
private int h1() {
  return f(); // This is fine.
}
private int h2() {
  return g(); // Compilation error. g() can throw, but h2() can't. The exception has nowhere to go.
}
private int h3() {
  try {
    return g(); // This is fine. We catch the potential exception, and h3() doesn't throw anything.
  } catch (IOException e) {
    return -1;
  }
}
private int h4() throws IOException {
  return g(); // This is fine. We don't catch the exception g() throws, so h4() will pass the exception along to its caller, which it's allowed to do.
}
```

For more detail, see https://docs.oracle.com/javase/tutorial/essential/exceptions/index.html.

jmchristen: Now, back to this specific case: `File.mkdir` doesn't throw `IOException` on failure, it reports failure via the return value. It's up to you to check that return value and handle the error how you want.

PR #24, opened 2020-08-18
Gcsbranch

freliciatucker:
- basically the same as PR #18 which i will delete, but much cleaner commit history


jmchristen:
Are you retracting PR #18 in favor of this one?

freliciatucker:
@jared yes. i havent deleted yet bc i wanted to look back to make sure all the changes were made

src/main/java/data/Database.java
```diff
@@ -1,79 +1,19 @@
...
 import java.io.OutputStream;
 import java.util.ArrayList;
 
-/* This class creates an instance of a database that a file can be written to and read from.
-Assumptions: for each runID, only 1 of each datatype can be associated with it. */
-public class Database {
-  /* Place where files are stored.*/
-  private String directory;
+public interface Database {
+  /* Abstract methods. */
```
jmchristen, line 9: You can document the intended function of the abstract methods here, on the interface. Then on the implementations of the interface, functions that override the abstract methods only need to document behavior specific to that implementation.

src/main/java/data/FileDatabase.java
```diff
@@ -0,0 +1,84 @@
...
+   */
+  @Override
+  public InputStream readData(String runID, String type) {
+    String fileName = fileName(runID, type);
+    try {
+      return new FileInputStream(fileName);
+    } catch (FileNotFoundException e) {
+      System.err.format("Unable to locate the give file with name %s", fileName);
+    }
+    return null;
```
jmchristen, line 42: You can move `return null` into the `catch` block to make the causality clear. Same below.

src/main/java/data/FileDatabase.java
```diff
@@ -0,0 +1,84 @@
...
+  private String fileName(String runId, String type) {
+    return getDatabaseName() + "/" + runId + "_" + type;
+  }
+
+  /* Returns list of files in database. */
+  @Override
+  public ArrayList<String> getAllFiles() {
+    File[] files = folder.listFiles();
+    ArrayList<String> filesAsStrings = new ArrayList<String>();
+    for (int i = 0; i < files.length; i++) filesAsStrings.add(files[i].getName());
```
jmchristen, line 70: Always use braces.

https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

jmchristen, line 70: I think `File.list` can do this for you.

https://docs.oracle.com/javase/8/docs/api/java/io/File.html#list--

src/main/java/shared/sharedObjects.java
```diff
@@ -9,7 +10,7 @@
   // public static
   static {
     try {
-      dataInstance = new Database("testdb-pose-3d-viewer-step");
+      dataInstance = new GCSDatabase("testdb-pose-3d-viewer-step");
```
jmchristen, line 13: This class looks out of date. See my comment on David's PR for why we shouldn't have a shared database instance: https://github.com/googleinterns/step199-2020/pull/16#discussion_r468225590.

freliciatucker: ok thanks. i read thru it and will need to read again bc still confused about thread safety. but if i change it the way david did in  his branch, then i would have to declare a database in each servlet and the servlets would know directly what type of database is being implemented within it bc i would initialize the database as new GCSDatabase(); . I thought the point of having a shared objects file was to keep the code modular and to remove the detail of which database is being used (and also to keep it simple if database implementation needs to update) . will ping to learn more

src/main/java/servlets/DataEntries.java
```diff
@@ -30,9 +32,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    // Logic that goes here that calls two methods from the intermediate Database
-    // depending on what exactly
-    String json = DatabaseQuery.getJson(sharedObjects.dataInstance);
-    response.getWriter().println(json);
+    // get JSON from Database.
+    try {
+      Database database = new GCSDatabase(sharedObjects.databaseName);
+      String json = DatabaseQuery.getJson(database);
+      response.getWriter().println(json);
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
freliciatucker: i reduced the code inside the try to just initializing the database (i had it like this before, but got compile time errors, but these errors arent shoeing anymore.) since it is only 1 line now in the try, does this meet the specific exception error message that you want?

src/main/java/servlets/DataEntries.java
```diff
@@ -30,9 +32,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      System.out.println("*************COULD NOT INITIALIZE DATABASE*********************");
+      System.out.println("Exception while initializing" + e.getMessage());
+      throw new RuntimeException(e.getMessage());
```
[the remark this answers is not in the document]
freliciatucker: are you suggesting replacing (Exception e) with (Throwable t)? (probably not bc this isnt encouraged, but am a lil confused.)
ooh or are you saying replace throw new... with throw e? imma go with that

src/main/java/servlets/DataEntries.java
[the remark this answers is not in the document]
jmchristen: This change did reduce the code inside `try` to only the statements that can throw, which is good. By "catch a more specific exception," I mean that you should rarely write `catch (Exception e)`, and instead write `catch (IOException e)` or whatever specific subclass of `Exception` can be thrown by the statement. `Exception` can otherwise catch a lot of different things that you might not really want to catch.

src/main/java/servlets/DataEntries.java
```diff
@@ -33,14 +35,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    try {
-      Database database = new GCSDatabase(sharedObjects.databaseName);
-      String json = DatabaseQuery.getJson(database);
-      response.getWriter().println(json);
-    } catch (Exception e) {
-      System.out.println("*************COULD NOT INITIALIZE DATABASE*********************");
-      System.out.println("Exception while initializing" + e.getMessage());
-      throw new RuntimeException(e.getMessage());
-    }
+    database = new GCSDatabase(sharedObjects.databaseName);
```
jmchristen, line 38: The `new GCSDatabase(...)` code will run on every call to `doGet`, so you're constructing a new database object on every GET request. I think you can override `GenericServlet.init` and do the construction there instead, so that you only create the database once instead of in every request.

https://javaee.github.io/javaee-spec/javadocs/javax/servlet/GenericServlet.html#init--

src/main/java/servlets/RunInfo.java
```diff
@@ -22,12 +25,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     System.out.println("The dataType is " + dataType);
-    DbReader dataReader = new DbReader(sharedObjects.dataInstance, runId, dataType);
-    Decoder.decode(dataReader.read(), response.getOutputStream());
+
     // We now make would make an instance of the reader object to get a stream from
     // the database, then pass this value to the decoder, with a string as our ouput
     // stream.
-    // As reader object is not yet defined instead get the file input stream
-    // directly
+    database = new GCSDatabase(sharedObjects.databaseName);
```
jmchristen, line 32: Same here, `new GCSDatabase` runs on every GET.

src/main/java/servlets/Upload.java
```diff
@@ -25,11 +29,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
     String dataType = "pose";
-    // Should take in database instance, along with data type.
-    DbWriter dataWriter = new DbWriter(sharedObjects.dataInstance, runid, dataType);
-    Part filePart = null;
 
-    filePart = request.getPart("file");
+    /*
+     * Make a database object and write file to it. redirect to index page.
+     */
+    database = new GCSDatabase(sharedObjects.databaseName);
```
jmchristen, line 36: Same here, `new GCSDatabase` runs on every POST.

jmchristen:
> I'm not sure exactly what Jared had in mind with his comment, but the google style guide I believe recommends not nesting throws in catches. We added this line when we were testing an error and were trying to get better debug information, but I don't think this should be in the final application. As all this does is catch and then rethrows the exception, you might as well just make the method throw the exception, I don't think this extra comment in general is very useful.

I can't find that recommendation in the style guide. But you are right that, rather than catch-and-rethrow, it usually makes more sense to simply not catch the exception in the first place and instead just let it propagate.

src/main/java/IO/DbWriter.java
```diff
@@ -1,72 +1,56 @@
...
-      /* In this case we need to try to generate a different runId. */
-      return getRandomRunId();
+  /* Gets an outputstream from the database. */
+  public OutputStream write() {
+    try {
+      return database.writeData(runId, type);
+    } catch (IOException e) {
     }
-    return newRunId;
+    return null;
```
jmchristen, line 53: You can move `return null` into the `catch` block to make the causality clear.

PR #25, opened 2020-08-24
Selection Pane Addition

D-moe:
Add support for selecting portions of runs, selection pane for multiple runs added (just the layout not the feature).
I forked this from the master branch which will not compile for java until Frelicia's branch merge fixes this problem so this error can be ignored.

### DIFF
--- a/src/main/java/proto/SensorData.java
+++ b/src/main/java/proto/SensorData.java
@@ -5,20 +5,21 @@ package proto;
 
 public final class SensorData {
   private SensorData() {}
-  public static void registerAllExtensions(
-      com.google.protobuf.ExtensionRegistryLite registry) {
+
+  public static void registerAllExtensions(com.google.protobuf.ExtensionRegistryLite registry) {}
+
+  public static void registerAllExtensions(com.google.protobuf.ExtensionRegistry registry) {
+    registerAllExtensions((com.google.protobuf.ExtensionRegistryLite) registry);
   }
 
-  public static void registerAllExtensions(
-      com.google.protobuf.ExtensionRegistry registry) {
-    registerAllExtensions(
-        (com.google.protobuf.ExtensionRegistryLite) registry);
-  }
-  public interface PoseOrBuilder extends
+  public interface PoseOrBuilder
+      extends
       // @@protoc_insertion_point(interface_extends:formats.Pose)
       com.google.protobuf.MessageOrBuilder {
 
     /**
+     *
+     *
      * <pre>
      * Position is in WGS84 latitude, longitude, altitude.
      * </pre>
@@ -28,6 +29,8 @@ public final class SensorData {
     double getGpsTimestamp();
 
     /**
+     *
+     *
      * <pre>
      * Degrees in [-90, 90].
      * </pre>
@@ -37,6 +40,8 @@ public final class SensorData {
     double getLat();
 
     /**
+     *
+     *
      * <pre>
      * Degrees in [-180,180].
      * </pre>
@@ -46,6 +51,8 @@ public final class SensorData {
     double getLng();
 
     /**
+     *
+     *
      * <pre>
      * Meters above WGS84
      * </pre>
@@ -55,6 +62,8 @@ public final class SensorData {
     double getAlt();
 
     /**
+     *
+     *
      * <pre>
      * Degrees in [0,360]
      * </pre>
@@ -64,6 +73,8 @@ public final class SensorData {
     float getRollDeg();
 
     /**
+     *
+     *
      * <pre>
      * Degrees in [0,360]
      * </pre>
@@ -73,6 +84,8 @@ public final class SensorData {
     float getPitchDeg();
 
     /**
+     *
+     *
      * <pre>
      * Degrees int [0,360]
      * </pre>
@@ -81,17 +94,16 @@ public final class SensorData {
      */
     float getYawDeg();
   }
-  /**
-   * Protobuf type {@code formats.Pose}
-   */
-  public  static final class Pose extends
-      com.google.protobuf.GeneratedMessageV3 implements
+  /** Protobuf type {@code formats.Pose} */
+  public static final class Pose extends com.google.protobuf.GeneratedMessageV3
+      implements
       // @@protoc_insertion_point(message_implements:formats.Pose)
       PoseOrBuilder {
     // Use Pose.newBuilder() to construct.
     private Pose(com.google.protobuf.GeneratedMessageV3.Builder<?> builder) {
       super(builder);
     }
+
     private Pose() {
       gpsTimestamp_ = 0D;
       lat_ = 0D;
@@ -103,10 +115,10 @@ public final class SensorData {
     }
 
     @java.lang.Override
-    public final com.google.protobuf.UnknownFieldSet
-    getUnknownFields() {
+    public final com.google.protobuf.UnknownFieldSet getUnknownFields() {
       return com.google.protobuf.UnknownFieldSet.getDefaultInstance();
     }
+
     private Pose(
         com.google.protobuf.CodedInputStream input,
         com.google.protobuf.ExtensionRegistryLite extensionRegistry)
@@ -121,60 +133,60 @@ public final class SensorData {
             case 0:
               done = true;
               break;
-            default: {
-              if (!input.skipField(tag)) {
-                done = true;
+            default:
+              {
+                if (!input.skipField(tag)) {
+                  done = true;
+                }
+                break;
               }
-              break;
-            }
-            case 9: {
-
-              gpsTimestamp_ = input.readDouble();
-              break;
-            }
-            case 17: {
-
-              lat_ = input.readDouble();
-              break;
-            }
-            case 25: {
-
-              lng_ = input.readDouble();
-              break;
-            }
-            case 33: {
-
-              alt_ = input.readDouble();
-              break;
-            }
-            case 45: {
-
-              rollDeg_ = input.readFloat();
-              break;
-            }
-            case 53: {
-
-              pitchDeg_ = input.readFloat();
-              break;
-            }
-            case 61: {
-
-              yawDeg_ = input.readFloat();
-              break;
-            }
+            case 9:
+              {
+                gpsTimestamp_ = input.readDouble();
+                break;
+              }
+            case 17:
+              {
+                lat_ = input.readDouble();
+                break;
+              }
+            case 25:
+              {
+                lng_ = input.readDouble();
+                break;
+              }
+            case 33:
+              {
+                alt_ = input.readDouble();
+                break;
+              }
+            case 45:
+              {
+                rollDeg_ = input.readFloat();
+                break;
+              }
+            case 53:
+              {
+                pitchDeg_ = input.readFloat();
+                break;
+              }
+            case 61:
+              {
+                yawDeg_ = input.readFloat();
+                break;
+              }
           }
         }
       } catch (com.google.protobuf.InvalidProtocolBufferException e) {
         throw e.setUnfinishedMessage(this);
       } catch (java.io.IOException e) {
-        throw new com.google.protobuf.InvalidProtocolBufferException(
-            e).setUnfinishedMessage(this);
+        throw new com.google.protobuf.InvalidProtocolBufferException(e).setUnfinishedMessage(this);
       } finally {
         makeExtensionsImmutable();
       }
     }
-    public static final com.google.protobuf.Descriptors.Descriptor
-        getDescriptor() {
+
+    public static final com.google.protobuf.Descriptors.Descriptor getDescriptor() {
       return proto.SensorData.internal_static_formats_Pose_descriptor;
     }
 
@@ -188,6 +200,8 @@ public final class SensorData {
     public static final int GPS_TIMESTAMP_FIELD_NUMBER = 1;
     private double gpsTimestamp_;
     /**
+     *
+     *
      * <pre>
      * Position is in WGS84 latitude, longitude, altitude.
      * </pre>
@@ -201,6 +215,8 @@ public final class SensorData {
     public static final int LAT_FIELD_NUMBER = 2;
     private double lat_;
     /**
+     *
+     *
      * <pre>
      * Degrees in [-90, 90].
      * </pre>
@@ -214,6 +230,8 @@ public final class SensorData {
     public static final int LNG_FIELD_NUMBER = 3;
     private double lng_;
     /**
+     *
+     *
      * <pre>
      * Degrees in [-180,180].
      * </pre>
@@ -227,6 +245,8 @@ public final class SensorData {
     public static final int ALT_FIELD_NUMBER = 4;
     private double alt_;
     /**
+     *
+     *
      * <pre>
      * Meters above WGS84
      * </pre>
@@ -240,6 +260,8 @@ public final class SensorData {
     public static final int ROLL_DEG_FIELD_NUMBER = 5;
     private float rollDeg_;
     /**
+     *
+     *
      * <pre>
      * Degrees in [0,360]
      * </pre>
@@ -253,6 +275,8 @@ public final class SensorData {
     public static final int PITCH_DEG_FIELD_NUMBER = 6;
     private float pitchDeg_;
     /**
+     *
+     *
      * <pre>
      * Degrees in [0,360]
      * </pre>
@@ -266,6 +290,8 @@ public final class SensorData {
     public static final int YAW_DEG_FIELD_NUMBER = 7;
     private float yawDeg_;
     /**
+     *
+     *
      * <pre>
      * Degrees int [0,360]
      * </pre>
@@ -277,6 +303,7 @@ public final class SensorData {
     }
 
     private byte memoizedIsInitialized = -1;
+
     public final boolean isInitialized() {
       byte isInitialized = memoizedIsInitialized;
       if (isInitialized == 1) return true;
@@ -286,8 +313,7 @@ public final class SensorData {
       return true;
     }
 
-    public void writeTo(com.google.protobuf.CodedOutputStream output)
-                        throws java.io.IOException {
+    public void writeTo(com.google.protobuf.CodedOutputStream output) throws java.io.IOException {
       if (gpsTimestamp_ != 0D) {
         output.writeDouble(1, gpsTimestamp_);
       }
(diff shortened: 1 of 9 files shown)


src/main/webapp/2Dscript.js
```diff
@@ -80,7 +151,279 @@ function addCenterControl(controlDiv) {
...
+      }
+
+    });
+    checkBoxEntry.appendChild(input);
+    currentRow.appendChild(checkBoxEntry);
+    table.appendChild(currentRow);
+    console.log(currentRow);
+    even = !even;
+
+  }
```
jmchristen, line 230: This loop is huge; please refactor into smaller functions.

src/main/webapp/2Dscript.js
```diff
@@ -60,12 +106,37 @@ function formatPoseData() {
...
+  const selectionUI = document.createElement('div');
+  selectionUI.className = 'selectionUI';
+  selectionUI.title = 'Select the pose runs to view/render in 3D.';
+  sideControlDiv.appendChild(selectionUI);
+
+  // Set CSS for the selectin pane interior.
+  const selectionText = document.createElement('div');
+  selectionText.className = 'selectionText';
+  selectionText.appendChild(innerContent);
+  // Need to call a function to create a table with checkboxes here, likely write some generic function to do it with any JSON file possible. Also add checkbox as column and an eyeball icon with event listeners.
```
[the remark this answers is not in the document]
D-moe: Yeah still need to resolve some of the linter errors, this is one of them

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +151,279 @@ function addCenterControl(controlDiv) {
...
+    table.appendChild(currentRow);
+    console.log(currentRow);
+    even = !even;
+
+  }
+  return table;
+
+}
+
+function plotLine(dataEntries) {
```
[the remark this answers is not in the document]
D-moe: ditto to linter error

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +151,279 @@ function addCenterControl(controlDiv) {
...
+        console.log("The value of dataentries is " + dataEntries);
+        // if the value is not found in the cache then fetch it.
+        if (dataEntries === undefined) {
+          console.log("Data Entries was false");
+          console.log("The id is:" + event.target.id);
+          fetch("/getrun?id=" + event.target.id + "&dataType=pose").then(response => response.json())
+            .then(data => dataEntries = data).then(() => {
+              datas[event.target.id] = dataEntries;
+              const toGraph = plotLine(dataEntries);
+              maps[event.target.id] = toGraph;
```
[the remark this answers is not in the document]
D-moe: What's the rationale for this change, is it more efficient or less error prone?

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +151,279 @@ function addCenterControl(controlDiv) {
 
   // Setup the click event listeners: simply set the map to Chicago.
   controlUI.addEventListener('click', function() {
-    window.location.href = '/3DVisual.html?id=' +
-    runId + '&dataType=' + dataType;
+    window.location.href = '/3DVisual.html?id=' + id + '&dataType=' + type;
   });
 }
+
+function createSideTable(json) {
```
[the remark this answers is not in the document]
D-moe: ditto linter

src/main/webapp/2Dscript.js
```diff
@@ -1,56 +1,102 @@
...
-  // Intitalizes the map.
+  // Occurs when google map api calls all promises, which is getting this
+  // function called somehow, according to stack trace.
+  if (pose === undefined) {
+    return;
+  }
+
+  console.log('initMap called');
+  const poseCoordinates = [];
+  for (let i = 0; i < pose.length; i++) {
```
jmchristen, line 56: Can you use `for (const point of pose)`?

https://google.github.io/styleguide/jsguide.html#features-for-loops

src/main/webapp/2Dscript.js
```diff
@@ -1,56 +1,102 @@
...
+// Define bounding box around the different data sets to quickly determined
+// whether they should be included in the calculation or not
 
+/**
+ * A Point object with a lat and lng.
+ * @typedef {Object<number, number>} Point
+ * @property {number} lat The lat of the point.
+ * @property {number} lng The lng of the point.
+ */
+let flightPath;
```
jmchristen, line 20: What is `flightPath`? The name is surprising to me.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +151,279 @@ function addCenterControl(controlDiv) {
...
+    strokeOpacity: 1.0,
+    strokeWeight: 2
+  });
+  flightPath.setMap(null);
+  console.log("The value of map is " + map);
+  return currentLineGraph;
+}
+
+// When finished drawing the box, instead of leaving the widget, create markers at the current lat and intersection with the line.
+$(function () {
```
jmchristen, line 255: This is another huge and deeply nested function, please refactor. I am not familiar with the `$` syntax, what is it doing?

src/main/webapp/2Dscript.js
D-moe: Yeah was working on that wanted to get it pushed though, so people could see what was going on.

src/main/webapp/2Dscript.js
D-moe: I used Jquery in this code to simplify the many nested event handlers $( gives a global selector in jquery

src/main/webapp/2Dscript.js
```diff
@@ -1,56 +1,102 @@
 // Gobal variables.
```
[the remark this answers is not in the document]
D-moe: My bad, should have been gobble like a turkey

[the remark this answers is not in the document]
D-moe: will correct

src/main/webapp/2Dscript.js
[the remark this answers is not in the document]
zjwoody: The way of setting a property using [] does not interact with the Map data structure. It uses the feature of the generic object. The value is not stored in the Map for queries. The operations like has/delete on the data will fail.
See details:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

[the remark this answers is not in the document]
D-moe: I guess I don't understand why we should use a map instead of an object in this case? As this is a simple string key pairing, it seems like these limitations are acceptable. Also, its possible we may want to convert this map object into a json string for passing through sessionstorage etc, a limitation that a map doesn't allow I believe? https://www.geeksforgeeks.org/map-vs-object-in-javascript/ (for reference on the JSON conversion issue)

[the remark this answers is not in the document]
D-moe: I guess perhaps the naming convention was confusing as I titled it maps as it stores the lines that are graphed not because it is a dataMap.

src/main/webapp/2Dscript.js
```diff
@@ -1,24 +1,43 @@
...
+ * from the requisite proto in the JSON format.
+ * Position is given in WGS84 latitude, longitude, and altitude.
+ * @typedef {Object<number, number, number, number,
+ * number, number, number>} PoseData
+ * @property {number} gpsTimestamp The time in seconds.
+ * @property {number} lat Degrees in [-90, 90]
+ * @property {number} lng Degrees in [-180, 180]
+ * @property {number} alt Meters above WGS84
+ * @property {number} rollDeg Degrees in [0,360]
+ * @property {number} pitchDeg Degrees in [0,360]
```
jmchristen, line 24: I think pitch is actually in [-90, 90].

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +169,361 @@ function addCenterControl(controlDiv) {
...
+ * @param {number} valLat
+ * @param {number} valLng
+ * @return {boolean}
+ */
+function withinBound(minLat, maxLat, minLng, maxLng, valLat, valLng) {
+  if ((valLat >= minLat && valLat <= maxLat) &&
+    (valLng >= minLng && valLng <= maxLng)) {
+    return true;
+  }
+  return false;
```
jmchristen, line 528: This can be simplified:
`return lat >= minLat && lat <= maxLat && lng >= minLng && lng <= maxLng;`

src/main/webapp/2Dscript.js
```diff
@@ -1,56 +1,122 @@
...
+const dataCache = {};
 let pose;
-let runId;
-let dataType;
+let data;
+let id;
+let type;
+let currentLat;
+let currentLong;
+let initialPoseData;
```
jmchristen, line 11: It's getting pretty difficult to follow the data flow with all the global variables. Can you make these into local variables passed around explicitly as much as possible?

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +169,361 @@ function addCenterControl(controlDiv) {
...
+
+
+/**
+ * Generate a polyline from the given data points and return it.
+ * @param {Array<Point>} dataEntries
+ * @return {google.maps.Polyline}
+ */
+function plotLine(dataEntries) {
+  const currentLine = [];
+  for (let i = 0; i < dataEntries.length; i++) {
```
jmchristen, line 337: Can you use `for (point of dataEntries)`?

src/main/webapp/3Dscript.js
```diff
@@ -208,3 +191,33 @@ function animate() {
...
+  const queryString = window.location.search;
+  const urlParams = new URLSearchParams(queryString);
+  const id = urlParams.get('id');
+  const type = urlParams.get('dataType');
+  // Only need to refetch the data if it is not contained in local storage, in
+  // general it should be.
+  const isStored = urlParams.get('stored');
+  const subSectionNumber = urlParams.get('subSection');
+  console.log(id + type);
+  if (!isStored) {
```
jmchristen, line 209: Prefer to test the positive condition first:
```
if (isStored) {
  ...
} else {
  ...
}
```
That way the reader doesn't have to do double negation.

https://drive.google.com/file/d/1-g637f-xAyGWbcILf5QbLRiTmvsDA3P9/view?usp=sharing

src/main/webapp/2Dscript.js
D-moe: Copied this from the master branch proto comment, so would that mean it was wrong there too?

src/main/webapp/2Dscript.js
D-moe: I'll work on moving some of these into functions. I know I was having a bit of a challenge figuring out the proper scope for variables passed to event handlers, any tips for that?

src/main/webapp/2Dscript.js
jmchristen: Probably so.

src/main/webapp/2Dscript.js
jmchristen: I'm not sure what you mean. Are you talking about something like:
```
onclick(event) {
  map.setSomething(event.something) // map is currently global
}
```
and wondering what the scope of `map` should be? Or something else?

D-moe: Yeah so it gets confusing to me if something like this occurs
function doSomething(){
let a;
let b;
b.addEventListener('click', function(){
// reference the a which is the scope of doSomething
console.log(a);
});
vs if you factor out this lambda into a function, passing that value to the function without using global scope. I tried using bind, but I think I was misunderstanding how to use 'this' in this case?

jmchristen: How about:
```
function doSomething() {
  let a;
  let b;
  b.addEventListener('click', function() { handleClick(a) });
}
function handleClick(a) {
  console.log(a);
}
```
There is still a lambda, but it's very short even if `handleClick` is long.

D-moe: for (int i =0 ; i <5; i++){
a.push(i)

D-moe: b[i].addEventListener

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +180,410 @@ function addCenterControl(controlDiv) {
...
+        position: latLng,
+      });
+      infoWindow.setMap(map);
+    },
+    );
+
+    subPath.setMap(map);
+
+    markerBottom = genMarker(discoveredMinLat, discoveredMinLatPair);
+    markerTop = genMarker(discoveredMaxLngPair, discoveredMaxLng);
```
jmchristen, line 462: It seems odd to reassign `markerTop` and `markerBottom` on each iteration of the loop. What's the purpose?

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +180,410 @@ function addCenterControl(controlDiv) {
...
+  console.log(pose);
+  const poseLength = pose.length;
+  // Lat can be from [-90,90].
+  let discoveredMinLat = 91;
+  let discoveredMinLatPair = 181;
+  // Lng can be from [-180,180].
+  let discoveredMaxLng = -181;
+  let discoveredMaxLngPair = -91;
+  subLine = [];
+  for (let i = 0; i < poseLength; i++) {
```
jmchristen, line 550: `for (point of pose)`

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +180,410 @@ function addCenterControl(controlDiv) {
...
+  $('.widget').remove();
+
+  const mapRuns = Object.entries(runs);
+  // Clear pop up window.
+  if (infoWindow !== undefined) {
+    infoWindow.setMap(null);
+  }
+  // Clear other paths.
+
+  for ([id, currentRun] of mapRuns) {
```
jmchristen, line 426: Can the loop body be factored out into a function?

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +180,410 @@ function addCenterControl(controlDiv) {
...
+      console.log('Polyline clicked at lat: ' + latLng.lat() +
+        ' lng: ' + latLng.lng());
+      infoWindow = new google.maps.InfoWindow({
+        content: '<a href=/home.html?id=' + id + '&dataType=' +
+          type + '&subSection=' + subSectionNumber +
+          '&stored=true' + '> View in 3D </a>',
+        position: latLng,
+      });
+      infoWindow.setMap(map);
+    },
```
jmchristen, line 456: Can the listener be factored out into a function?

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +180,410 @@ function addCenterControl(controlDiv) {
...
+  return polyLine;
+}
+
+/**
+ * Compute the subsection of a line contained in a bounding rectangle.
+ * @param {Point} pose
+ * @param {number} currentLat
+ * @param {number} priorLat
+ * @param {number} currentLng
+ * @param {number} priorLng
```
jmchristen, line 533: I think `pose` is actually an Array? And current/prior is surprising terminology for a bounding rectangle; I would expect something like min/max.

src/main/webapp/2Dscript.js
D-moe: In this case it doesn't really make sense to have the loop, this is a remnant for trying to add an additional feature, I'll remove.

src/main/webapp/2Dscript.js
D-moe: Yeah should be Array<Point> good catch. The reasoning for this was the first click gives the priorLat and priorLng and the next click gives the currentLng and currentLng that you clicked to. That's why I used that terminology as I was thinking in terms of when the events happen not the bounds of the rectangle itself.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +175,418 @@ function addCenterControl(controlDiv) {
...
+/**
+ * Generate a polyline from the given data points and return it.
+ * @param {Array<Point>} dataEntries
+ * @return {google.maps.Polyline}
+ */
+function plotLine(dataEntries) {
+  const currentLine = [];
+  for (point of dataEntries) {
+    currentLine.push({lat: point.lat, lng: point.lng});
+  }
```
jmchristen, line 343: I think you can do
`const currentLine = [...dataEntries];`
or
`const currentLine = dataEntries.slice();`
instead.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +175,418 @@ function addCenterControl(controlDiv) {
...
+  // Clear pop up window.
+  if (infoWindow !== undefined) {
+    infoWindow.setMap(null);
+  }
+  // Clear other paths.
+  // line
+  // minLatPoint
+  // lat , lng
+  // minLngPoint
+  // lat, lng
```
jmchristen, line 422: Extraneous comment.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +175,418 @@ function addCenterControl(controlDiv) {
...
+ * Create a polyline from the given points with specified parameters.
+ * @param {Point} linePoints
+ * @param {string} color
+ * @param {number} opacity
+ * @param {number} weight
+ * @param {number} index
+ * @return {google.maps.Polyline}
+ */
+function getPolyLine(linePoints, color, opacity, weight, index = 1) {
+  const polyLine = new google.maps.Polyline({
```
jmchristen, line 498: Follow the API's lead on capitalization: "polyline" is all one word, so the L shouldn't be capitalized.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +175,418 @@ function addCenterControl(controlDiv) {
...
+/**
+ * Compute the subsection of a line contained in a bounding rectangle.
+ * @param {Array<Point>} pose
+ * @param {number} currentLat
+ * @param {number} priorLat
+ * @param {number} currentLng
+ * @param {number} priorLng
+ * @return {subSectionData}
+ */
+function computeSubSection(pose, currentLat, priorLat, currentLng, priorLng) {
```
jmchristen, line 523: I think `currentLat, currentLng, priorLat, priorLng` would be a better order. We're conditioned to see lat,lng pairs together.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +175,418 @@ function addCenterControl(controlDiv) {
...
+  // Lat can be from [-90,90].
+  let discoveredMinLat = 91;
+  let discoveredMinLatPair = 181;
+  // Lng can be from [-180,180].
+  let discoveredMaxLng = -181;
+  let discoveredMaxLngPair = -91;
+  subLine = [];
+  for (point of pose) {
+    const loopLat = point.lat;
+    const loopLng = point.lng;
```
jmchristen, line 538: I don't think these variable carry their weight. Just refer to `point.lat` and `point.lng` within the loop, so the reader doesn't have to keep another two variables in mind.

src/main/webapp/2Dscript.js
```diff
@@ -80,7 +175,418 @@ function addCenterControl(controlDiv) {
...
+ * @param {number} valLat
+ * @param {number} valLng
+ * @return {boolean}
+ */
+function withinBound(minLat, maxLat, minLng, maxLng, valLat, valLng) {
+  if (valLat >= minLat && valLat <= maxLat &&
+    valLng >= minLng && valLng <= maxLng) {
+    return true;
+  }
+  return false;
```
jmchristen, line 591: This can be simplified:
`return lat >= minLat && lat <= maxLat && lng >= minLng && lng <= maxLng;`

src/main/webapp/2Dscript.js
D-moe: This was a mistake in the type annotation, where dataEntries is of type PoseData not Point and this loop removes those extra parameters.

D-moe: I fixed this error in the type annotation

src/main/webapp/2Dscript.js
```diff
@@ -1,76 +1,171 @@
...
+
+  initialPoseMap = getPolyline(formatPoseData(initialPose), '#FF0000', 1.0, 2);
+  initialPoseMap.setMap(map);
+
+  const centerControlDiv = centerControl();
   map.controls[google.maps.ControlPosition.TOP_CENTER].push(centerControlDiv);
+
+  // Add event listeners to record latitude and longitude globally for selection
+  // box.
+  map.addListener('mousemove', function(event) {
```
freliciatucker, line 107: just a question: is there any benefits for mousemove over click events?

src/main/webapp/2Dscript.js
```diff
@@ -1,76 +1,171 @@
...
+  // function called somehow, according to stack trace.
+  const initialPose = runs[id].data;
+  const initialPoseMap = runs[id].map;
+  if (initialPose === undefined) {
+    return;
+  }
+
   map = new google.maps.Map(document.getElementById('map'), {
-    center: {lat: pose[0].lat, lng: pose[0].lng},
+    center: {lat: initialPose[0].lat, lng: initialPose[0].lng},
```
freliciatucker, line 95: could also display timestamp with initalPose[i].gpsTimeStamp

src/main/java/proto/SensorData.java
```diff
@@ -5,20 +5,21 @@
...
   private SensorData() {}
-  public static void registerAllExtensions(
-      com.google.protobuf.ExtensionRegistryLite registry) {
-  }
 
-  public static void registerAllExtensions(
-      com.google.protobuf.ExtensionRegistry registry) {
-    registerAllExtensions(
-        (com.google.protobuf.ExtensionRegistryLite) registry);
+  public static void registerAllExtensions(com.google.protobuf.ExtensionRegistryLite registry) {}
```
jmchristen, line 9: This file's changes seem unrelated, can you split into another PR?

D-moe: This is the result of mvn spotless:apply which Jie mentioned to pass the Java check. This issue arose because some of the code in master (apparently these lines) wasn't in the appropriate format. I think it's fine for everyone to run mvn spotless:apply and commit even if it changes this file multiple times as git will recognize the change as the same and we should likely diligently run the command anyways and that way all checks will pass so we can prevent accidentally putting code in master that is in the improper format.

PR #26, opened 2020-08-24
Converting Coordinate Systems

batldre:


src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
+
+/**
+ * Finds the median value of the latitude, longitude,
+ * and altitude. The median is used as a reference point
+ * for the East-North-Up (ENU) conversion and as the center
+ * point for the 2D map.
+ * @return {array} An array containing median lat, lng, and alt
+ *   coordinates.
+ */
+function findMedians() {
```
jmchristen, line 75: Why not just choose an arbitrary point as the ENU origin? It's simpler, and you don't know anything about the shape of the trajectory, so it's hard to say whether the median will always be a better choice.

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
 
   // Add the light to the scene.
-  let light = new THREE.PointLight(0xffffff, 1, 0 );
-  light.position.set(0, 100, 0);
-  scene.add(light)
+  const light = new THREE.PointLight(0xffffff, 1, 0 );
+  light.position.set(0, 0, 1000);
+  scene.add(light);
+
+  createInstances();
```
freliciatucker, line 121: it doesnt really make sense to me why add map would need to call these functions.  could there be a funciton in fetch data that calls them? seems a bit more logic friendly, tho i see it doesnt change the code fundamentally

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
-  scene.add(orientation);
   plotOrientation();
 }
 
+
 /**
- * This uses the pose data to create a blue line representing the pose
- * trajectory. 
+ * Creates all the children for each axis of orientation. Each object
+ * is a geometry instance, initializing the object with one geometry
```
freliciatucker, line 129: is each object referring to the entire mesh or like each individual cylinder/point? wondering bc of click on pose implementation considerations

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
  * Creates the 2D terrain and points a light at it.
- * It also calls two other functions so it can wait on the fetch 
- * request and significantly cut down rendering time with 
+ * It also calls two other functions so it can wait on the fetch
+ * request and significantly cut down rendering time with
  * geometry instancing.
 */
 function addMap() {
+  const [medianLat, medianLng, medianAlt] = findMedians();
+  median = {lat: medianLat, lng: medianLng, alt: medianAlt};
```
jmchristen, line 101: Maybe you can define a Point class to make it easier to pass around this type of data?

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
-  for (point of pose){
+  for (const axis in orientation) {
+    if (axis != '') {
+      matrixRotation(axis);
+      // Allows the GUI to manipulate the orientation.
+      orientation[axis].instanceMatrix.needsUpdate = true;
+      orientation[axis].position.x = poseTransform.translateX;
+      orientation[axis].position.y = poseTransform.translateY;
+      orientation[axis].position.z = poseTransform.translateZ;
+      orientation[axis].rotation.y = THREE.Math.degToRad(poseTransform.rotateY);
```
freliciatucker, line 261: is there no need to rotate on the x axis?

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
+
 /**
- * This uses the pose data to create a blue line representing the pose
- * trajectory. 
+ * Creates all the children for each axis of orientation. Each object
+ * is a geometry instance, initializing the object with one geometry
+ * call rather than the number of data points. (O(1) vs. O(N))
+ */
+function createInstances() {
+  // Blue corresponds to the Z axis.
```
jmchristen, line 133: All three comments refer to "the Z axis."

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
+/**
+  * Converts latitude, longitude, and altitude coordinates (LLA) to
+  * Earth-Centered, Earth-Fixed coordinates (ECEF).
+  * @param {double} lat Latitude represented in Decimal Degrees
+  * @param {double} lng Longitude represented in Decimal Degrees
+  * @param {double} alt Altitude represented in Decimal Degrees
+  * @return {array} An array containing x, y, z ECEF coordinates.
+  */
+function llaToEcef(lat, lng, alt) {
+  // Constants.
```
jmchristen, line 161: Include units.

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
+    Math.pow(minorAxis, 2)) / Math.pow(majorAxis, 2);
+  const lambda = THREE.Math.degToRad(lat);
+  const phi = THREE.Math.degToRad(lng);
+  const radius = majorAxis /
+    Math.sqrt(1 - eSq * Math.pow(Math.sin(lambda), 2));
+  // Conversion.
+  const x = (radius + alt) * Math.cos(lambda) * Math.cos(phi);
+  const y = (radius + alt) * Math.cos(lambda) * Math.sin(phi);
+  const z = ((Math.pow(minorAxis, 2) * radius) /
+    Math.pow(majorAxis, 2) + alt) * Math.sin(lambda);
```
jmchristen, line 176: Can you add a link to the equation you're using?

src/main/webapp/3Dscript.js
```diff
@@ -1,29 +1,32 @@
-import {FlyControls} from 'https://threejs.org/examples/jsm/controls/FlyControls.js';
+// import {FlyControls} from 'https://threejs.org/examples/jsm/controls/FlyControls.js';
```
freliciatucker, line 1: should comment be removed?

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
+
+  return [x, y, z];
+}
+
+/**
+ * Converts Earth-Centered, Earth-Fixed coordinates (ECEF) to
+ * East-North-Up coordinates (ENU).
+ * @param {double} x Latitude represented in ECEF coordinates.
+ * @param {double} y Longitude represented in ECEF coordinate.
+ * @param {double} z Altitude represented in ECEF coordinates.
```
jmchristen, line 186: ECEF does not have lat/lng/alt. It's just x/y/z (probably in meters). An increase in x, for example, can correspond to an increase in altitude in one place on the globe, a decrease in altitude on the opposite side of the globe, and neither elsewhere.

src/main/webapp/3Dscript.js
```diff
@@ -45,144 +49,285 @@ function initThreeJs() {
...
+ *     ECEF coordinates.
+ * @param {double} lon0 The reference longitude represented in
+ *     ECEF coordinates.
+ * @param {double} h0 The reference height represented in
+ *     ECEF coordinates.
+ * @return {array} An array containing x, y, z ENU coordinates.
+ */
+function ecefToEnu(x, y, z, lat0, lon0, h0) {
+  // Converts LLA reference point to ECEF coordinates.
+  const [x0, y0, z0] = llaToEcef(lat0, lon0, h0);
```
jmchristen, line 197: I think you'll end up doing a lot of repeated conversions of your origin point from LLA to ECEF. You could convert it once and pass it to this function already in ECEF.

src/main/webapp/3Dscript.js
zjwoody: +1
For the client code that not be able to add unit test, you can test it manually to make sure the code can run and doesn't contain bugs.

src/main/webapp/3Dscript.js
zjwoody: +1
Or you can simply get the four corners of the trajectory instead sort(since it is 2D map, you don't need alt): north east, north west, south east and south west, and then you can compute the center point easily.

src/main/webapp/3Dscript.js
batldre: Each object refers to the 3 instanced meshes which each have thousands of colored cylinders. Each point now looks something like this -> https://i.stack.imgur.com/6ySFH.png

src/main/webapp/3Dscript.js
batldre: Don't see why, any rotation on the x axis usually clips the pose through the 2D map. Also, the ENU coordinate system should always keep the pose parallel to the ground.

src/main/webapp/3Dscript.js
batldre: @jmchristen I used the median mainly because of alignment issues with the 2D map. The Maps Api requires a center point and the ENU conversion requires a reference origin point, the median is the quickest way I could think to align the two. The Maps Static Api also doesn't allow you to specific radius, so while I find a way to generate large 3D terrain, I used median to cover as many points as possible. 
@zjwoody  How would you suggest doing this for a line? I made this as a temporary solution until I could replace the map with terrain, would the center point be a better option?

src/main/webapp/3Dscript.js
batldre: Do you mean units for the constants in the function? I wrote the lat, long, and alt coordinates were in decimal degrees above.

src/main/webapp/3Dscript.js
D-moe: Can you use the proper types for JSDoc comments, while double is a more exact description of the type, it is not a valid JSDoc type which means we can't run static type checking using these comments? Instead you can put @param {number} and specify that it is type double in the comment. That way we can have static type linting to prevent as many runtime bugs. Not a huge deal either way, just thought I'd mention it.
https://jsdoc.app/tags-type.html

D-moe:
Looks like this repo is not running the linter checks, an issue we saw before.

D-moe:
Not sure where this branch was generated from but inside the github workflows config the linter should be set to run on push not pull_request, was how I resolved this issue before. I thought this was pushed to master but maybe not?

D-moe:
Looks like it was, maybe also add that change here in a commit then just so the checks will run before merge.

src/main/webapp/3Dscript.js
jmchristen: If the radius is small enough to be a problem in practice, then this is okay. (If it's not a problem in practice, I would simplify by just choosing the first point.) My first thought would be to get the center like Jie suggested, but I can't actually think of a strong reason to prefer it over your implementation, other than it's easier to explain.

src/main/webapp/3Dscript.js
jmchristen: Yes, for the constants (those which aren't dimensionless, of course).

jmchristen: Oh, and for the return values.

src/main/webapp/3Dscript.js
```diff
@@ -1,7 +1,32 @@
...
+  /**
+   * A getter.
+   * @param {String} axis A string specifying which axis to return;
+   * @return {(number|object)} An arbitary 3D data point.
+   */
+  get(axis) {
+    if (axis == 'x') return this.x;
+    if (axis == 'y') return this.y;
+    if (axis == 'z') return this.z;
+  }
```
jmchristen, line 28: Wouldn't it be easier just to access `point.x` directly? (It's possible a typedef or record would be more appropriate than a class; I'm not totally sure about JS best practices. But this pattern of translating string to property looks suspect.)

src/main/webapp/3Dscript.js
```diff
@@ -45,146 +65,273 @@ function initThreeJs() {
...
+  const loader = new THREE.TextureLoader();
+  const material = new THREE.MeshLambertMaterial({
     map: loader.load(
-      'https://maps.googleapis.com/maps/api/staticmap' +
-      '?format=png&center=' + pose[0].lat + ',' + pose[0].lng + 
-      '&zoom=18&size=500x500&key=' + apiKey)
-    });
-  let geometry = new THREE.PlaneGeometry(50,50);
+        'https://maps.googleapis.com/maps/api/staticmap' +
+      '?format=png&center=' + center.get(x) + ',' + center.get(y) +
```
jmchristen, line 130: LLA points should use lat/lng/alt, not x/y/z.

src/main/webapp/3Dscript.js
```diff
@@ -45,146 +65,273 @@ function initThreeJs() {
...
-      (point.alt - pose[0].alt)/4,
-      (point.lat - pose[0].lat) * 25000 * -poseTransform.scale)
-    );
+  const coordinates=[];
+  const [ecefX, ecefY, ecefZ] =
+    llaToEcef(origin.get(x), origin.get(y), origin.get(z));
+  const ecefOrigin = new Point(ecefX, ecefY, ecefZ);
+  for (const point of pose) {
+    const [x0, y0, z0] = llaToEcef(point.lat, point.lng, point.alt);
+    const [x, y, z] = ecefToEnu(x0, y0, z0, ecefOrigin, origin);
```
jmchristen, line 260: `Point` can come in handy here.

src/main/webapp/3Dscript.js
```diff
@@ -45,146 +76,272 @@ function initThreeJs() {
...
+  const queryString = window.location.search;
+  const urlParams = new URLSearchParams(queryString);
+  const id = urlParams.get('id');
+  const type = urlParams.get('dataType');
+  fetch('/getrun?id=' + id + '&dataType=' + type)
+      .then((response) => response.json())
+      .then((data) => pose = data)
+      .then(() => {
+        const [lat, lng, alt] = findMedians();
+        const lla = {latitude, longitude};
```
jmchristen, line 93: The variable doesn't include altitude, so maybe it should be named `latLng`.

src/main/webapp/3Dscript.js
```diff
@@ -45,146 +76,272 @@ function initThreeJs() {
...
-  geometry = new THREE.CylinderBufferGeometry(.005, .005, .1);
-  orientation = new THREE.InstancedMesh(geometry, material, pose.length);
-  orientation.instanceMatrix.setUsage(THREE.DynamicDrawUsage);
-  scene.add(orientation);
-  plotOrientation();
+  const xCylinder = new THREE.InstancedMesh(geometry, material, pose.length);
+  xCylinder.instanceMatrix.setUsage(THREE.DynamicDrawUsage);
+  scene.add(xCylinder);
+
+  return new Point(xCylinder, yCylinder, zCylinder);
```
jmchristen, line 183: This seems like an abuse of `Point`. I wouldn't expect a point to have anything other than scalar components.

src/main/webapp/3Dscript.js
zjwoody: e.g. you have an array of 2D(lat, lng) points, then you can easily find a bounding box of these points, so the box has four corner, it is easy to get the center = ((northest_lat+southest_lat)/2, (eastest_lng+westest_lng)/2). If it is a line(it means two corners are overlapped with the other two), you can still get the correct center.

src/main/webapp/3Dscript.js
```diff
@@ -45,146 +76,273 @@ function initThreeJs() {
...
+ *     of the 3D world.
+ * @param {Object} origin Reference LLA position, used as the origin
+ *     of the 3D world.
+ * @return {array} An array containing x, y, z ENU coordinates.
  */
-function plotTrajectory() {
+function ecefToEnu(ecefPose, ecefOrigin, origin) {
+  // ECEF to ENU equation: https://www.mathworks.com/help/map/ref/ecef2enu.html
+  // Localizes the new xyz coordinate using the reference point.
+  const xd = ecefPose.getX() - ecefOrigin.getX();
```
zjwoody, line 232: local_x

src/main/webapp/3Dscript.js
```diff
@@ -45,146 +76,273 @@ function initThreeJs() {
...
+  const cosPhi = Math.cos(THREE.Math.degToRad(origin.getY()));
+  const sinPhi = Math.sin(THREE.Math.degToRad(origin.getY()));
+
+  // Matrix multiplication.
+  const xEast = -sinPhi * xd + cosPhi * yd;
+  const yNorth = -cosPhi * sinLambda * xd - sinLambda *
+    sinPhi * yd + cosLambda * zd;
+  const zUp = cosLambda * cosPhi * xd + cosLambda *
+    sinPhi * yd + sinLambda * zd;
+
```
zjwoody, line 247: Could you add a link to the math reference?

batldre: https://www.mathworks.com/help/map/ref/ecef2enu.html, this link is given above on line 230.

src/main/webapp/3Dscript.js
```diff
@@ -57,143 +88,261 @@ function fetchData() {
...
-      '?format=png&center=' + pose[0].lat + ',' + pose[0].lng +
+      '?format=png&center=' + center.latitude + ',' + center.longitude +
       '&zoom=18&size=500x500&key=' + apiKey),
   });
-  let geometry = new THREE.PlaneGeometry(50, 50);
+  const geometry = new THREE.PlaneGeometry(200, 200);
   const map = new THREE.Mesh(geometry, material);
-  map.position.set(0, -4, 0);
-  map.rotation.x = -Math.PI / 2;
+  map.position.set(0, 0, -4);
```
freliciatucker, line 148: maybe put magic numbers in variables?

src/main/webapp/3Dscript.js
```diff
@@ -57,143 +88,261 @@ function fetchData() {
...
+/**
+ * Creates all the children for each axis of orientation. Each object
+ * is a geometry instance, initializing the object with one geometry
+ * call rather than the number of data points. (O(1) vs. O(N))
+ * @return {object} A Point object holding a cylinder for each axis.
+ */
+function createInstances() {
+  // Blue corresponds to the Z axis.
+  let material = new THREE.MeshBasicMaterial({color: 'blue'});
+  const geometry = new THREE.CylinderBufferGeometry(.005, .005, .1);
```
freliciatucker, line 167: magic numbers here too

src/main/webapp/3Dscript.js
```diff
@@ -57,143 +88,261 @@ function fetchData() {
...
+ */
+function findMedians() {
+  const arrays = {lat: [], lng: [], alt: []};
+  for (const point of pose) {
+    arrays.lat.push(point.lat);
+    arrays.lng.push(point.lng);
+    arrays.alt.push(point.alt);
+  }
+  const sorted = {
+    lat: [...arrays.lat].sort((a, b) => a - b),
```
freliciatucker, line 119: just a question: this notation looks cool. what does it do?

src/main/webapp/3Dscript.js
batldre: Could I maybe put a comment explaining? Adding 3 variables for each axis seems kind of redundant.

src/main/webapp/3Dscript.js
batldre: Same with these magic numbers, can I instead add a comment?

src/main/webapp/3Dscript.js
jmchristen: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

batldre: Sorts the array given using the compare function given as a parameter. The a - b specifies that it should be in ascending order. Ill add a comment for this too.

src/main/webapp/3Dscript.js
```diff
@@ -44,140 +94,340 @@ function initThreeJs() {
...
-   * increment in the 6th decimal place equates to a 0.025 unit
-   * change in our 3D space. Since altitude is already represented in
-   * meters, we simply divide by 4 to adjust for our 1:4 unit ratio.
-   */
-  const x = (lng - pose[0].lng) * 25000 * poseTransform.scale;
-  const y = (alt - pose[0].alt) / 4;
-  const z = (lat - pose[0].lat) * 25000 * -poseTransform.scale;
-  return [x, y, z];
+function ecefToEnu(ecefPose, ecefOrigin, origin) {
+  // ECEF to ENU equation: https://www.mathworks.com/help/map/ref/ecef2enu.html
```
zjwoody, line 274: I didn't find the equation, could you point out where the equation is?

batldre: Sorry, I accidentally pasted another website I was referencing. http://www.hydrometronics.com/downloads/Ellipsoidal%20Orthographic%20Projection.pdf 
This is the real link, the matrix algebra is on page 15.

zjwoody (approved):

batldre:
Merged code has been transfered to the Merge David Andre Branch pull request

PR #28, opened 2020-08-26
adds checkboxes to index page

freliciatucker:


src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
...
+      currentRow.appendChild(columnEntry);
     }
-    currentRow.appendChild(columnEntry);
     index++;
   }
   return currentRow;
 }
+
+/* Select all Checkboxes function. */
+function toggle(source) {
```
freliciatucker, line 78: toggle and json called in index page which is already merged to master with changes. todo: delete makemultirunurl since it is replaced by json function

src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
...
+ * Make url for multiple runs selected. should be in the style of
+ ** link.href = "/home.html?id=" + key + "&dataType=" + json[key]
+ */
+function makeMultiRunURL() {
+  const checkboxes = document.getElementsByName('check');
+  let count =0;
+  const link = document.createElement('a');
+  link.href = '/home.html?';
+
+  for (let i=0, n=checkboxes.length; i<n; i++) {
```
D-moe, line 95: Probably just separate this out into a function, also use for(box of checkboxes).

D-moe: Also you create the checkboxes so you shouldn't have to query the dom to retrieve their values. Consider storing them in a map/ list upon creation, so you can access them directly.

src/main/webapp/script.js
```diff
@@ -13,8 +13,11 @@ function createTable() { // eslint-disable-line no-unused-vars
     columnOne.innerText = 'RunID';
```
D-moe, line 13: If you look at my current PR I condensed this into a function and an array to make it a lot more concise and less redundant, consider that.

src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
   let index = 0;
   for (const type of json[key]) {
     const columnEntry = document.createElement('td');
-    console.log(type);
-    console.log(type.toUpperCase());
+
     if ((type.toUpperCase() === 'POSE' && index === 0) ||
       (type.toUpperCase() === 'POINTCLOUD' && index === 1)) {
       const link = document.createElement('a');
```
D-moe, line 57: Can you split this into a function as well?

src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
...
-    currentRow.appendChild(columnEntry);
     index++;
   }
   return currentRow;
 }
+
+/* Select all Checkboxes function. */
+function toggle(source) {
+  checkboxes = document.getElementsByName('check');
+  for (let i=0, n=checkboxes.length; i<n; i++) {
```
jmchristen, line 80: `for (const checkbox of checkboxes)`?

src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
...
+  const link = document.createElement('a');
+  link.href = '/home.html?';
+
+  for (let i=0, n=checkboxes.length; i<n; i++) {
+    if (checkboxes[i].checked) {
+      count++;
+      console.log('here what we are splitting!' +
+       checkboxes[i].getAttribute('value'));
+      const splitInfo = checkboxes[i].getAttribute('value').split('_');
+      link.href += 'id' + count + '=' + splitInfo[0] + '&type' +
```
jmchristen, line 101: `+=` on a string within a loop is always a red flag, because the string has to be copied on each iteration. You could accumulate substrings within an array, then use `Array.join` to concatenate them all outside of the loop.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
...
+      count + '=' + splitInfo[1]+ '-';
```
jmchristen, line 102: Consider template literals for complex string construction like this.

https://google.github.io/styleguide/jsguide.html#features-strings-template-strings

src/main/webapp/script.js
```diff
@@ -47,22 +51,80 @@ function initializeRow(json, key, currentRow) {
...
+function makeMultiRunJson() {
+  const checkboxes = document.getElementsByName('check');
+  let str= '{';
+  for (let i=0, n=checkboxes.length; i<n; i++) {
+    if (checkboxes[i].checked) {
+      console.log('here what we are splitting!' +
+      checkboxes[i].getAttribute('value'));
+      const splitInfo = checkboxes[i].getAttribute('value').split('_');
+      str += '\n' + '"' + splitInfo[0] + '": [' + '\n "' +
+      splitInfo[1] + '" \n ], \n';
```
jmchristen, line 120: Same string comments here.

src/main/webapp/script.js
```diff
@@ -36,6 +39,9 @@ function createTable() { // eslint-disable-line no-unused-vars
     }
   });
 }
+
+const checkMap = new Map();
```
jmchristen, line 43: You never use the keys when accessing elements of `checkMap`. Should it just be an `Array`?

src/main/webapp/script.js
zjwoody: +1.
Also give a meaningful name for the column, e.g. runIdColumn instead columnOne, etc.

src/main/webapp/script.js
```diff
@@ -13,8 +16,11 @@ function createTable() { // eslint-disable-line no-unused-vars
...
     const columnTwo = document.createElement('th');
     columnTwo.innerText = 'DataType';
+    const columnThree = document.createElement('th');
+    columnThree.innerText = 'Select';
     headerRow.appendChild(columnOne);
     headerRow.appendChild(columnTwo);
+    headerRow.appendChild(columnThree);
     table.appendChild(headerRow);
     let even = false;
     for (const key in json) {
```
zjwoody, line 26: Instead of "json", you can give it a more representative name, e.g. runInformation. 
Same as other names. Always use meaningful name to avoid confusion.

src/main/webapp/script.js
```diff
@@ -47,22 +55,73 @@ function initializeRow(json, key, currentRow) {
   let index = 0;
   for (const type of json[key]) {
     const columnEntry = document.createElement('td');
```
zjwoody, line 57: Do you need to check whether the json data has this key?
Also rename the parameter names to make it more meaningful.

src/main/webapp/script.js
```diff
@@ -36,6 +42,8 @@ function createTable() { // eslint-disable-line no-unused-vars
     }
   });
 }
+
+
 /**
  * Create a row in the given table.
```
zjwoody, line 48: Seems the function should name appendRow

src/main/webapp/script.js
freliciatucker: this comment is might be best answered by @D-moe  since its his code, but i can go ahead and give the variables/functions better names

D-moe: Eslint requires a check to determine if the json data has a key. I was surprised by this too but it seems that it is possible to get inside the loop without this condition being true.

src/main/webapp/script.js
```diff
@@ -29,40 +35,93 @@ function createTable() { // eslint-disable-line no-unused-vars
...
+  currentRow.appendChild(checkbox);
+  checkArray.push(checkbox);
+}
+
+/* eslint-disable */
+/**
+* Select all Checkboxes function.
+ * @param {object} source checkbox
+ */
+function toggle(source) {
```
jmchristen, line 97: "toggle" suggests it will invert the state of the boxes, but I think this actually sets them to match the input box?

freliciatucker: in coding examples i found online for checking checkboxes, they used the name toggle to describe the same action as what i do in this function: 
https://www.developphp.com/video/JavaScript/Checkbox-Toggle-Tutorial-Select-Deselect-All-Elements
https://stackoverflow.com/questions/37286629/toggle-checkbox-using-javascript

This is the first time I'm learning of the word, so i wanted to use it and mirror how other people used it in practice, but can change name to checkAllBoxes

PR #32, opened 2020-08-26
2D multiple run

D-moe:
Implement multiple runs in 3D viewer, modify GUI features, fully implement selection of multiple runs in 2D viewer.
It is expected that the java check should fail as the master branch did not compile when this branch (and the one it is merging into) were created.


src/main/webapp/2Dscript.js
```diff
@@ -4,6 +4,10 @@
 let map;
 // Store all information for every run (acts as cache), of @type{Run}
 const runs = {};
+/*
+runId:{map data subsection checkBox Element colorElement
+}
+*/
```
jmchristen, line 10: Extraneous comment.

src/main/webapp/2Dscript.js
```diff
@@ -202,19 +208,77 @@ function createSideTable(json) {
...
+      const columnElements = [keyEntry, checkBoxEntry, colorPickerEntry];
       updateRow(currentRow, columnElements);
       table.appendChild(currentRow);
+
+      table.appendChild(currentRow);
+      console.log(currentRow);
       // Toggle the value of even for every other row.
       even = !even;
     }
   }
```
jmchristen, line 279: This loop got huge and deeply nested; see if you can factor out some functions.

src/main/webapp/2Dscript.js
```diff
@@ -202,19 +208,77 @@ function createSideTable(json) {
...
+          runs[runId].map !== undefined) {
+          // Set the new center to be the first latlng value fetched from the
+          // data.
+          const mapData = runs[runId].data[0];
+          map.setCenter({lat: mapData.lat, lng: mapData.lng});
+        } else {
+          // In this case the box has become unchecked. We want to remove this
+          // graph. We can only remove this box if it has been generated before
+          // and then can be removed appropriately.
+          const toRemove = runs[event.target.id].map;
```
jmchristen, line 258: Why `event.target.id` here but `runId` above?

src/main/webapp/2Dscript.js
```diff
@@ -414,40 +476,67 @@ function placeRectangleEnd() {
...
-  if (markerTop !== undefined) {
-    markerTop.setMap(null);
+  clearSelectedPaths(['subSection', 'markerBottom', 'markerTop']);
+  // Clear other paths. line minLatPoint lat , lng minLngPoint lat, lng Iterate
+  // over all data values here, need to find a way to check the checkbox value,
+  // should likely store this in an object somewhere to prevent excess DOM
+  // queries.
+  const mapRuns = Object.entries(runs);
+  for ([id, currentRun] of mapRuns) {
+    if (currentRun.checkBox.checked) {
```
jmchristen, line 486: You can invert the conditional to reduce nesting:
```
if (!currentRun.checkBox.checked) {
  continue;
}
/* ... nominal-case code ... */
```
https://docs.google.com/document/d/16SSi-rIUGxJ5-ksifO6y0InCOwgbRa3lHBLnBJ8Klrk/edit?usp=sharing

src/main/webapp/2Dscript.js
```diff
@@ -414,40 +476,67 @@ function placeRectangleEnd() {
...
+      // clicked.
+      google.maps.event.addListener(subPath, 'click', linkTo3D);
+      subPath.setMap(map);
+      currentRun.markerBottom = genMarker(subSectionData.minLatPoint.lat,
+          subSectionData.minLatPoint.lng);
+      currentRun.markerTop = genMarker(subSectionData.maxLngPoint.lat,
+          subSectionData.maxLngPoint.lng);
+      showAll([currentRun.subSection, currentRun.markerBottom,
+        currentRun.markerTop]);
+      isMouseDown = false;
```
jmchristen, line 502: Can you move this line out into the outer event handler? The causality is more clear there. Same for `isMouseDown = true` in `placeRectangleStart`.

src/main/webapp/2Dscript.js
```diff
@@ -414,40 +476,67 @@ function placeRectangleEnd() {
...
+    }
   }
-  if (subPath !== undefined) {
-    subPath.setMap(null);
+  const subSectionObject = {};
+  for ([id, currentRun] of mapRuns) {
+    if (currentRun.checkBox.checked) {
```
jmchristen, line 507: You can invert the conditional to reduce nesting.

src/main/webapp/2Dscript.js
```diff
@@ -414,40 +476,67 @@ function placeRectangleEnd() {
...
+                  JSON.stringify(subSectionObject));
+              window.location.href = 'home.html?subsection=true';
+            });
+            infoWindow = new google.maps.InfoWindow({
+              content: link,
+              position: {lat: latLng.lat(), lng: latLng.lng()},
+            });
+            console.log(infoWindow);
+            infoWindow.setMap(map);
+          });
```
jmchristen, line 537: Can you move this lambda out into a named function?

src/main/webapp/2Dscript.js
```diff
@@ -505,6 +594,23 @@ function getPolyLine(linePoints, color, opacity, weight, index = 1) {
...
 }
+/**
+ * Clear prior markers/path if they exist.
+ * @param {Array<google.maps.Polyline | google.maps.InfoWindow |
+ * google.maps.Marker>} pathArray
+ */
+function clearSelectedPaths(pathArray) {
+  const objects = Object.values(runs);
+  for (val of objects) {
+    // Now take all these
```
jmchristen, line 605: Unfinished comment?

src/main/webapp/2Dscript.js
```diff
@@ -505,6 +594,23 @@ function getPolyLine(linePoints, color, opacity, weight, index = 1) {
   });
   return polyLine;
 }
+/**
+ * Clear prior markers/path if they exist.
+ * @param {Array<google.maps.Polyline | google.maps.InfoWindow |
+ * google.maps.Marker>} pathArray
```
jmchristen, line 600: Is it okay for the type info to break across lines like that? It looks odd with Github's syntax highlighting.

src/main/webapp/3Dscript.js
```diff
@@ -1,31 +1,49 @@
-// import {FlyControls} from 'https://threejs.org/examples/jsm/controls/FlyControls.js';
 import {OrbitControls} from 'https://threejs.org/examples/jsm/controls/OrbitControls.js';
 import {GUI} from 'https://threejs.org/examples/jsm/libs/dat.gui.module.js';
-
+/**
+* A Point object with a lat and lng.
+* @typedef {Object<number, number, number>} Point
+* @property {number} lat The lat of the point.
+* @property {number} lng The lng of the point.
+* @property {number} alt The alt of the point.
```
jmchristen, line 8: Document the units (degrees, meters above WGS84).

src/main/webapp/3Dscript.js
```diff
@@ -1,31 +1,49 @@
...
+*/
 // These are global objects.
+/**
+ * @typedef {Array<number, number>}
```
jmchristen, line 12: What does it mean for an Array to have multiple parameters? And does this mean `runs` is an array, but also has the properties `data` and `color`?

src/main/webapp/3Dscript.js
```diff
@@ -1,31 +1,49 @@
...
+ * @property {Array<Point>} data All the stored run data for a given runId
+ * @property {string} color The color in a standard HTML acceptable format
+ */
+let runs;
+/*
+runId: data color
+
+*/
```
jmchristen, line 20: Extraneous comment.

src/main/webapp/3Dscript.js
```diff
@@ -128,17 +152,18 @@ function plotTrajectory() {
...
-     * of the pose orientation locally instead using the world transform.
-     * This means it will rotate relative to the position of the matrix,
-     * not the center of the 3D scene.
+     * Matrix is a 4x4 matrix used to translate and rotate each instance of the
+     * pose orientation locally instead using the world transform. This means it
+     * will rotate relative to the position of the matrix, not the center of the
+     * 3D scene.
      */
     const matrix = new THREE.Matrix4();
     const [x, y, z] = llaDegreeToLocal(point.lat, point.lng, point.alt);
```
jmchristen, line 169: I think this needs to be updated to `llaDegreeToLocal`'s new signature.

src/main/webapp/3Dscript.js
```diff
@@ -43,58 +61,63 @@ function initThreeJs() {
...
   plotOrientation();
 }
 
 /**
  * Converts lla coordinates to local world coordinates.
- * @param {double} lat Latitude in degrees.
- * @param {double} lng Longitude in degrees.
- * @param {double} alt Altitude in meters.
- * @return {array} An array returning gps coordinates.
+ * @param {Array<Point>} pose The run to center the data around.
```
jmchristen, line 108: I think it makes more sense to accept a single point as the origin, and let the callers use `pose[0]` as necessary.

src/main/webapp/3Dscript.js
```diff
@@ -193,33 +229,32 @@ function animate() {
...
-    addMap();
+
+  const isSubSection = urlParams.get('subsection');
+  console.log('The value of subsection is ' + isSubSection);
+  let firstData;
+  if (isSubSection) {
+    runs = JSON.parse(sessionStorage.getItem('subsection'));
+    firstData = Object.values(data)[0].data;
+    // set the runId to We should loop through all the runs and call
+    // addPoseData() for all of them.
```
jmchristen, line 245: Mangled comment?

src/main/webapp/3Dscript.js
D-moe: It should be Array<Object<number,number>>. Will resolve this

src/main/webapp/2Dscript.js
D-moe: It highlights it properly in vsCode, believe it is acceptable as there's also that 80 line word limit.

src/main/webapp/2Dscript.js
```diff
@@ -414,40 +462,68 @@ function placeRectangleEnd() {
...
+    if (currentRun.checkBox.checked) {
+      const subSectionData = computeSubSection(currentRun.data,
+          currentLat, priorLat, currentLng, priorLng);
+      currentRun.subData = subSectionData.subData;
+      // Clear prior paths, only display newly selected ones.
+      const subPath = getPolyLine(currentRun.subData, 'blue', 1.0, 2, 1000);
+      // Setup event listener to show option for 3D window when polyline is
+      // clicked.
+      currentRun.subSection = subPath;
+      // google.maps.event.addListener(subPath, 'click', linkTo3D);
```
freliciatucker, line 481: can take out comment if not planning on bringing it back to life

src/main/webapp/2Dscript.js
```diff
@@ -414,40 +462,68 @@ function placeRectangleEnd() {
...
+        currentRun.markerTop]);
+      isMouseDown = false;
+    }
   }
-  if (subPath !== undefined) {
-    subPath.setMap(null);
+  const subSectionObject = {};
+  for ([id, currentRun] of mapRuns) {
+    if (currentRun.checkBox.checked && currentRun.subData.length !== 0) {
+      subSectionObject[id] = {};
```
freliciatucker, line 495: seems like a long section. could probably refactor into smaller functions

src/main/webapp/3Dscript.js
```diff
@@ -152,38 +189,58 @@ function plotOrientation() {
...
+      .onFinishChange(() => plotTrajectory(currentId.value, currentData))
+      .name('X Axis Translation');
+  gui.add(transforms[currentId.value], 'translateZ', -10, 10, .025)
+      .onChange(() => plotOrientation(currentId.value))
+      .onFinishChange(() => plotTrajectory(currentId.value, currentData))
+      .name('Z Axis Translation');
+  gui.add(transforms[currentId.value], 'scale', .5, 2, .25)
+      .onChange(() => plotOrientation(currentId.value))
+      .onFinishChange(() => plotTrajectory(currentId.value, currentData))
+      .name('Pose Scale Multiplier');
```
freliciatucker, line 237: question us three can discuss later, but should we have multiple guis? rn its alot going on it. and we dont want it to be too complicated.

D-moe: I agree this is getting complicated, I'm not totally sure what our approach should be or whether it is possible to create multiple dat.gui instances on one page. We certainly could try it so our display doesn't get too cluttered. Luckily this code only adds one additional text box to select the run so not too much clutter when actually rendered.

src/main/webapp/2Dscript.js
```diff
@@ -292,9 +295,84 @@ function generateCheckBoxEntry(runId) {
...
+  viewIcon.id = 'view_' + runId;
+  viewIcon.innerHTML = '<i class=\'fa fa-eye\'></i>';
+  viewIcon.addEventListener('click', function() {
+    const viewId = this.id;  // eslint-disable-line
+    const runId = viewId.split('_')[1];
+    console.log('run id is ' + runId);
+    const isChecked = runs[runId].checkBox.checked;
+    console.log('The checkbox value is ' + isChecked);
+    if (isChecked && runs[runId].data !== undefined &&
+      runs[runId].map !== undefined) {
```
jmchristen, line 359: Maybe you could pull this logic out into a helper (and re-use it, e.g. in `changeRunColor`):
```
function runIsVisible(run) {
  return run.checkBox.checked && run.data !== undefined && run.map !== undefined;
}
```
Even one-liner helper functions can be beneficial. In this case it relieves the reader from having to piece together the meaning of three different comparisons in order to understand the purpose of the conditional.

src/main/webapp/2Dscript.js
```diff
@@ -414,42 +489,73 @@ function placeRectangleEnd() {
...
-  const subSectionNumber = 1;
-  // Clear prior paths, only display newly selected ones.
-  if (markerBottom !== undefined) {
-    markerBottom.setMap(null);
+  clearSelectedPaths(['subSection', 'markerBottom', 'markerTop']);
+  // Clear other paths. line minLatPoint lat , lng minLngPoint lat, lng Iterate
+  // over all data values here, need to find a way to check the checkbox value,
+  // should likely store this in an object somewhere to prevent excess DOM
+  // queries.
+  const mapRuns = Object.entries(runs);
```
jmchristen, line 497: I think this demonstrates how a `Map` can improve over an object: it's more natural to iterate over an `Map`. You don't need to change it in this PR though.

src/main/webapp/2Dscript.js
```diff
@@ -414,42 +489,73 @@ function placeRectangleEnd() {
...
+  for ([id, currentRun] of mapRuns) {
+    if (!(currentRun.checkBox.checked && currentRun.subData.length !== 0)) {
+      continue;
+    }
+    subSectionObject[id] = {};
+    subSectionObject[id].color = currentRun.color.value;
+    subSectionObject[id].data = currentRun.subData;
+    console.log(subSectionObject);
+    google.maps.event.addListener(currentRun.subSection,
+        'click', (event) => createInfoWindow(event, subSectionObject));
```
jmchristen, line 528: Why loop through twice instead of doing it all together? (It may help to pull each of the current loop bodies into a helper function if you do collapse them.)

src/main/java/proto/SensorData.java
```diff
@@ -1,1026 +0,0 @@
-// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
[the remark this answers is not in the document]
D-moe: This file is autogenerated every time the program is compiled so it doesn't make sense to include it in the git history. It was accidentally added early on and is easy to continue adding so I added it to the .gitignore and also removed it from the cache so it wouldn't repeatedly add itself on git add -u.

src/main/webapp/2Dscript.js
```diff
@@ -292,9 +295,84 @@ function generateCheckBoxEntry(runId) {
...
+ * map elements.
+ * @param {string} runId
+ * @return {Array<HTMLElement, HTMLElement>}
+ */
+function generateColorPickerEntry(runId) {
+  const colorPickerEntry = document.createElement('td');
+  const colorPicker = document.createElement('input');
+  colorPicker.type = 'color';
+  colorPicker.id = 'color_' + runId;
+  // Initialize each value to a random starting color.
```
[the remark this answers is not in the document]
D-moe: Interesting, didn't really think about this nonlinear mapping.

src/main/webapp/2Dscript.js
D-moe: Good to know, thanks for the example.

src/main/webapp/2Dscript.js
```diff
@@ -495,40 +495,60 @@ function placeRectangleEnd() {
...
+/**
+ * Grab the selected region and generate the necessary polylines and markers.
+ * @param {Run} currentRun
+ * @param {number} currentLat
+ * @param {number} priorLat
+ * @param {number} currentLng
+ * @param {number} priorLng
+ */
+function generatedSelectedRegion(currentRun, currentLat, priorLat,
+    currentLng, priorLng) {
```
jmchristen, line 521: Prefer to keep these ordered as (lat, lng) pairs.

src/main/webapp/3Dscript.js
```diff
@@ -43,105 +57,119 @@ function initThreeJs() {
...
- * @return {array} An array returning gps coordinates.
+ * @param {number} runId The given run's id.
+ * @param {Point} pose The run to center the data around.
+ * @param {number} lat Latitude in degrees.
+ * @param {number} lng Longitude in degrees.
+ * @param {number} alt Altitude in meters.
+ * @return {Array<number, number, number>} An array returning gps coordinates.
  */
-function llaDegreeToLocal(lat, lng, alt) {
+function llaDegreeToLocal(runId, pose, lat, lng, alt) {
```
zjwoody, line 116: You can rename the "pose" to "reference_pose", or "origin" to make it more obviously to user.

src/main/webapp/3Dscript.js
```diff
@@ -1,31 +1,45 @@
...
 let renderer;
 let clock;
 let controls;
-let orientation;
-let trajectory;
-let pose;
-const poseTransform = {
+
+// Likely need to make a separate pose transform for each runId;
+const transforms = {};
```
zjwoody, line 25: You can rename it runIdToTransform that can easily know it is a map from runId to its transform.

Good name can make your code have better readability.

src/main/webapp/3Dscript.js
zjwoody: Try to give a descriptive name to make your code have better readability.
E.g. runIdToData.

PR #34, opened 2020-09-01
Fix master

D-moe:
Resolve some merge issues, minor bug fixes.

### DIFF
--- a/src/main/webapp/2Dscript.js
+++ b/src/main/webapp/2Dscript.js
@@ -237,7 +237,6 @@ function showPoseData(event) {
   if (checked) { // eslint-disable-line
     // First check the cache for this value.
     dataEntries = runs[id].data;
-    console.log('The value of dataentries is ' + dataEntries);
     // If the value is not found in the cache then fetch it.
     if (dataEntries === undefined) {
       fetchAndGraphData(id);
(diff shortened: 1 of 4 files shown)


src/main/webapp/2Dscript.js
```diff
@@ -658,12 +658,12 @@ function clearSelectedPaths(pathArray) {
  * Compute the subsection of a line contained in a bounding rectangle.
  * @param {Array<Point>} pose
  * @param {number} currentLat
- * @param {number} currentLng
  * @param {number} priorLat
+ * @param {number} currentLng
  * @param {number} priorLng
  * @return {subSectionData}
  */
-function computeSubSection(pose, currentLat, currentLng, priorLat, priorLng) {
```
jmchristen, line 666: This was correct as-is. We prefer to keep arguments ordered as (lat, lng) pairs since the concept is so common.

src/main/webapp/2Dscript.js
```diff
@@ -677,15 +677,17 @@ function computeSubSection(pose, currentLat, currentLng, priorLat, priorLng) {
   let discoveredMaxLngPair = -91;
   subData = [];
   for (point of pose) {
+    const loopLat = point.lat;
+    const loopLng = point.lng;
     if (withinBound(minLat, maxLat, minLng, maxLng, loopLat, loopLng)) {
```
jmchristen, line 682: This was correct without the extra variables (I don't think they carry their weight). You can instead change from `loopLat` => `point.lat`, etc.

src/main/webapp/2Dscript.js
```diff
@@ -716,13 +718,13 @@ function linkTo3D(event, subSectionObject) { // eslint-disable-line
  * @param {number} maxLat
  * @param {number} minLng
  * @param {number} maxLng
- * @param {number} lat
- * @param {number} lng
+ * @param {number} valLat
+ * @param {number} valLng
  * @return {boolean}
  */
-function withinBound(minLat, maxLat, minLng, maxLng, lat, lng) {
```
jmchristen, line 723: `lat, lng` is fine here.

PR #36, opened 2020-09-03
Merging David Andre Branch

batldre:


.gitignore
```diff
@@ -1,3 +1,2 @@
 target
 .stylelintrc.json
-src/main/java/proto/SensorData.java
```
jmchristen, line 3: You can sync this line so the big generated file gets ignored.

jmchristen (commented):
Looks like this undoes a bunch of intentional changes we made in code review of other PRs, am I missing something?

zjwoody:
I'd recommend not doing merge between your local branches, first check in pr1 to master, then sync pr2 with master, fix the conflict and then check in pr2.

zjwoody:
Use a meaningful pr name to describe what the pr is doing, so it will be easy to track the history of what feature is checked in which pr.

PR #37, opened 2020-09-03
Terrain

D-moe:
Implementation for the terrain feature; this is still a work in progress and likely shouldn't be merged to master due to long load times (7+ seconds blocked by the python script). TODO items include: properly extract lat long from structure, scale pose data appropriately (gridSize as well as altitude height), figure out why the python script returns such a large array of data, as it is much larger than one would expect for a 30m DEM scan.

### DIFF
--- a/src/main/java/python/testFinalish.py
+++ b/src/main/java/python/testFinalish.py
@@ -1,0 +1,76 @@
+import ee
+import json
+import sys
+import numpy as np
+
+ee.Initialize()
+expectedNumArgs = 4
+if(len(sys.argv) != expectedNumArgs + 1):
+    print("Wrong number of args provided")
+    exit(-1)
+argList = sys.argv
+#print(argList[0])
+minLat = float(argList[1])
+minLng = float(argList[2])
+maxLat = float(argList[3])
+maxLng = float(argList[4])
+area = ee.Geometry.Polygon([[minLat, minLng],\
+                            [maxLat, minLng],\
+                            [maxLat, maxLng],\
+                            [minLat, maxLng],\
+                            [minLat, minLng]])
+
+def LatLonImg(img):
+    img = img.addBands(ee.Image.pixelLonLat())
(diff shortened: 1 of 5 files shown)


zjwoody:
Thanks David working on this challenge feature!
After fixing the lint error, we can check in it.

zjwoody (approved):

src/main/java/python/testFinalish.py
```diff
@@ -0,0 +1,76 @@
...
+                counter+=1
+                arr[len(uniqueLats)-1-y,x] = data[counter] # we start from lower left corner
+    return arr
+
+#dataset = ee.Image('USGS/GMTED2010').clip(area)
+dataset = ee.Image("JAXA/ALOS/AW3D30/V2_2").clip(area)
+lat, lon, data = LatLonImg(dataset)
+ 
+# 1d to 2d array
+image  = toImage(lat,lon,data)
```
zjwoody, line 71: Seems image is unused.

src/main/java/python/testFinalish.py
```diff
@@ -0,0 +1,76 @@
...
+def LatLonImg(img):
+    img = img.addBands(ee.Image.pixelLonLat())
+ 
+    img = img.reduceRegion(reducer=ee.Reducer.toList(),\
+                                        geometry=area,\
+                                        scale=100)
+ 
+    data = np.array((ee.Array(img.get("AVE_DSM")).getInfo()))
+    lats = np.array((ee.Array(img.get("latitude")).getInfo()))
+    lons = np.array((ee.Array(img.get("longitude")).getInfo()))
```
zjwoody, line 32: I think you convert from image => list => array too many times, try to think a way to avoid the similar structure convert.

zjwoody (changes requested):

D-moe:
@googlebot I signed it!

